PR #6, opened 2017-07-23
STITCH-529 Fix continueTask method of StitchTasks in iOS SDK

jsflax:
Made StitchTask into an actual Promise system, not just a pseudo-callback system.

### DIFF
--- a/StitchCore/StitchCore/Source/Network/StitchTask.swift
+++ b/StitchCore/StitchCore/Source/Network/StitchTask.swift
@@ -1,105 +1,5 @@
 import Foundation
 
-/// Promise system to handle requests to Stitch
-open class StitchTask<Result> {
-    
-    private let queue: OperationQueue
-    
-    /// The final result of this task
-    public var result: StitchResult<Result>? {
-        didSet{
-            queue.isSuspended = false
-        }
-    }
-    
-    // MARK: - Init
-    
-    /// Create a new `StitchTask`
-    public init() {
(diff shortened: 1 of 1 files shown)


jsflax:
@mpobrien PTAL

StitchCore/StitchCore/Source/Network/StitchTask.swift
```diff
@@ -1,129 +1,337 @@
...
+    func execute(_ work: @escaping () -> Void)
+}
+
+extension DispatchQueue: ExecutionContext {
+    public func execute(_ work: @escaping () -> Void) {
+        self.async(execute: work)
+    }
+}
+
+public final class InvalidatableQueue: ExecutionContext {
```
mpobrien: question from our slack convo: this doesn't seem to be depended on anywhere in the code, maybe we don't need it?

jsflax: It can be used as an `ExecutionContext` by the consumer of the SDK. There are unused methods in both mobile SDKs with the intent that a consumer may use them. 

As long as we are doing our own promise impl, it might be a good idea to keep around.

jsflax:
There are unfortunately no tests for the Core module yet. There is a ticket for this that will be addressed sooner than later.

We should 100% have test coverage here though.

mpobrien:
@jsflax  understood, LGTM 👍

PR #9, opened 2017-10-09
Refer to Xcode's "Debug Area", not Android's "Android Manager"

ashitaka1:


jsflax:
Good catch– must've slipped through the cracks.

PR #11, opened 2017-10-19
STITCH-547 proactively refresh access tokens to avoid handling invalid session

adamchel:


jsflax:
This LGTM. I'll fix the formatting once the linter is added.

PR #20, opened 2017-12-12
STITCH-1009 iOS SDK doesn't persist StitchClient.auth

jsflax:


StitchCore/StitchCore/Source/Core/StitchClient.swift
```diff
@@ -111,6 +111,16 @@ public class StitchClient: StitchClientType {
...
     }
 
+    public var auth: Auth? {
+        if _auth == nil && isAuthenticated, let userId = self.httpClient.authInfo?.userId {
+            _auth = Auth(stitchClient: self,
+                         stitchHttpClient: self.httpClient,
+                         userId: userId)
+        }
+
+        return _auth
```
edaniels: [q] better to return nil?

jsflax: it's the same effect

jsflax:
The PR title is the ticket title: I'll squash the commit though

PR #21, opened 2017-12-14
STITCH-1016 - Fix broken email registration endpoint

adamchel:
`register` on the the `StitchClient` currently returns a 404, removing the trailing slash in the URL seems to fix it.


edaniels:
Please put on a JIRA ticket number in commit and in PR title before squash merging.

PR #22, opened 2017-12-19
STITCH-1014 resetPassword() method is missing password parameter in mobile SDKs

jsflax:


edaniels:
Did you test that doing reset password works now?

jsflax:
Yes, I have a local test that is passing, but it exposes my personal email. I will add a test once we have the admin SDK.

PR #23, opened 2017-12-19
STITCH-1012 Documentation of iOS SDK still uses pipelines instead of functions

jsflax:


kay-kim:
Not to butt in, but should we update the versions in lines 235-240?
https://github.com/jsflax/stitch-ios-sdk/blob/c7e532b93d66661b900ac061315b4ac748be1107/docs/index.html#L235-L240

jsflax:
@kay-kim Changed!

PR #24, opened 2017-12-19
STITCH-1022 Replace iOS StitchTask with PromiseKit

jsflax:
PromiseKit: https://github.com/mxcl/PromiseKit

PromiseKit seems to be the most used Promise library out there. It has been updated within the past 9 days, and is starred by over ~8k people, versus Bolts' ~1k. It also seems to be more idiomatic Swift.

I'm fairly certain there was a bug leftover in `StitchTask`. Now that we are using the new `PromiseKit` library, all tests are passing, so this also covers `STITCH-1020`.

PromiseKit has been added as a submodule.

NOTE: This is a breaking change.


edaniels:
@adamchel if you wanna take a look as well

MongoDBService/MongoDBService/Source/Core/Collection.swift
```diff
@@ -40,7 +41,7 @@ public struct Collection {
     @discardableResult
     public func find(query: Document,
                      projection: Document? = nil,
-                     limit: Int) -> StitchTask<[Document]> {
+                     limit: Int) -> Promise<[Document]> {
```
edaniels: So we're gonna need another major change right?

jsflax: Correct

PR #25, opened 2017-12-21
3.0.0 Changelog

jsflax:
Removed `StitchTask` in favor of `Promise` via the PromiseKit library: https://github.com/mxcl/PromiseKit/

Fixed bug in `auth` that prevented a full logout. `stitchClient.isAuthenticated` will now return properly.


edaniels:
Shouldn't the changes made be in separate commits referring to tickets or do those commits already exist?

PR #26, opened 2018-01-02
fixed an issue with decoding and already decoded xjson

messi:
Problem occurs on every findOne/InsertOne and probably all other calls as well.


messi:
Sorry, wrong branch

PR #29, opened 2018-01-18
iOS SDK: Namespace local storage of auth info by client app ID

jsflax:
Note: This does not include the migration. If we wish to include the migration, please update the JIRA ticket.

### DIFF
--- a/StitchCore/StitchCore/Source/Core/StitchClient.swift
+++ b/StitchCore/StitchCore/Source/Core/StitchClient.swift
@@ -5,11 +5,13 @@ import StitchLogger
 import Security
 import PromiseKit
 
-internal struct Consts {
+public struct Consts {
     static let ApiPath = "/api/client/v2.0/"
 
     //User Defaults
     static let UserDefaultsName = "com.mongodb.stitch.sdk.UserDefaults"
+
+    public static let defaultServerUrl = "https://stitch.mongodb.com"
 
     //keys
     static let ErrorKey = "error"
@@ -27,14 +29,14 @@ internal protocol StitchClientFactoryProtocol {
 public final class StitchClientFactory: StitchClientFactoryProtocol {
     typealias TClient = StitchClient
 
-    static func create(appId: String,
-                       baseUrl: String = "https://stitch.mongodb.com",
-                       networkAdapter: NetworkAdapter = StitchNetworkAdapter(),
-                       storage: Storage? = nil) -> Promise<StitchClient> {
+    public static func create(appId: String,
+                              baseUrl: String = Consts.defaultServerUrl,
+                              networkAdapter: NetworkAdapter = StitchNetworkAdapter(),
+                              storage: Storage? = nil) -> Promise<StitchClient> {
         return Promise(value: StitchClient.init(appId: appId,
-                                                baseUrl: baseUrl,
-                                                networkAdapter: networkAdapter,
-                                                storage: storage))
+                                                 baseUrl: baseUrl,
+                                                 networkAdapter: networkAdapter,
+                                                 storage: storage))
     }
 }
 
@@ -182,12 +184,12 @@ public class StitchClient: StitchClientType {
             guard let userDefaults = UserDefaults.init(suiteName: suiteName) else {
                 self.storage = MemoryStorage.init(suiteName: suiteName)!
                 printLog(.warning, text: "Invalid suiteName: \(suiteName)")
-                printLog(.warning, text: "Defaulting to memory storage. NOTE: App will not persist data in this state")
+                printLog(.warning, text: "Defaulting to memory storage. NOTE: App will not persist authentication status")
                 return
             }
             self.storage = userDefaults
             #else
-            printLog(.warning, text: "Defaulting to memory storage. NOTE: App will not persist data in this state")
+            printLog(.warning, text: "Defaulting to memory storage. NOTE: App will not persist authentication status")
             self.storage = MemoryStorage.init(suiteName: suiteName)!
             #endif
         }
--- a/StitchCore/StitchCore/Source/Core/StitchClient.swift
+++ b/StitchCore/StitchCore/Source/Core/StitchClient.swift
@@ -31,11 +31,12 @@ public class StitchClient: StitchClientType {
     internal var baseUrl: String
     internal let networkAdapter: NetworkAdapter
 
-    internal let userDefaults = UserDefaults(suiteName: Consts.UserDefaultsName)
+    internal var storage: Storage
 
     internal lazy var httpClient = StitchHTTPClient(baseUrl: baseUrl,
-                                                    appId: appId,
-                                                    networkAdapter: networkAdapter)
+                                                     appId: appId,
+                                                     networkAdapter: networkAdapter,
+                                                     storage: storage)
     private var authProvider: AuthProvider?
     private var authDelegates = [AuthDelegate?]()
 
@@ -87,7 +88,7 @@ public class StitchClient: StitchClientType {
         didSet {
             if let refreshToken = httpClient.authInfo?.refreshToken {
                 // save auth persistently
-                userDefaults?.set(true, forKey: Consts.IsLoggedInUDKey)
+                storage.set(true, forKey: Consts.IsLoggedInUDKey)
 
                 do {
                     let jsonData = try JSONEncoder().encode(httpClient.authInfo)
@@ -107,7 +108,7 @@ public class StitchClient: StitchClientType {
             } else {
                 // remove from keychain
                 try? self.httpClient.deleteToken(withKey: Consts.AuthJwtKey)
-                userDefaults?.set(false, forKey: Consts.IsLoggedInUDKey)
+                storage.set(false, forKey: Consts.IsLoggedInUDKey)
             }
         }
     }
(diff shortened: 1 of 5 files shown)


StitchCore/StitchCore/Source/Core/Storage.swift
```diff
@@ -0,0 +1,58 @@
+import Foundation
+
+/// Storage allows us to store arbitrary data via
+/// an arbitrary source. The protocol is modeled
+/// after \(Foundation.UserDefaults).
+public protocol Storage {
```
edaniels: Yay, not async!

edaniels:
Note: We will be including an auto migration

jsflax:
@edaniels @adamchel I've altered the original PR enough that I think this is worth a second look. PTAL

StitchCore/StitchCore/Source/Core/StitchClient.swift
```diff
@@ -168,11 +167,32 @@ public class StitchClient: StitchClientType {
...
         self.appId = appId
         self.baseUrl = baseUrl
         self.networkAdapter = networkAdapter
+
+        let suiteName = "\(Consts.UserDefaultsName).\(appId)"
+        if let storage = storage {
+            self.storage = storage
+        } else {
+            #if !os(Linux)
+            guard let userDefaults = UserDefaults.init(suiteName: suiteName) else {
```
adamchel: [style q] I saw this in several places, is there any reason we use the explicit `.init()` call rather than just calling the init function like `UserDefaults(suiteName: blah)`?

StitchCore/StitchCoreTests/AuthTests.swift
```diff
@@ -124,16 +124,19 @@ class AuthTests: XCTestCase {
...
         let exp = expectation(description: "multiple logins work as expected")
-        let mlsStitchClient = StitchClient(appId: "stitch-tests-ios-sdk-jjmum")
         var anonUserId = ""
         var emailUserId = ""
-        
-        // check storage
-        XCTAssertFalse(mlsStitchClient.isAuthenticated)
-        XCTAssertNil(mlsStitchClient.loggedInProviderType)
-        
+        var mlsStitchClient: StitchClient!
```
adamchel: Not sure why these changes are showing up here. This was stuff from my multiple login semantics commit that made it conform to the factory changes. Can we make sure we rebase everything properly before merging?

StitchCore/StitchCoreTests/StitchCoreTests.swift
```diff
@@ -25,6 +25,48 @@ class StitchCoreTests: XCTestCase {
     }
 
 
+    func testMultipleClientStorage() throws {
+        let exp = expectation(description: "multiple clients should store separately")
+
+        let creators = [StitchClientFactory.create(appId: "test1"),
+                        StitchClientFactory.create(appId: "test2")]
+
+        when(resolved: creators).done { (results: [Result<StitchClient>]) in
```
adamchel: fancy 😮

StitchCore/StitchCoreTests/StitchCoreTests.swift
```diff
@@ -25,6 +25,48 @@ class StitchCoreTests: XCTestCase {
...
+            XCTAssertEqual(sc1.storage.value(forKey: "fizz") as? String, "foo")
+            XCTAssertEqual(sc2.storage.value(forKey: "fizz") as? String, "bar")
+            exp.fulfill()
+        }
+
+        wait(for: [exp], timeout: 10)
+    }
+
+    func testMigrationLogic() throws {
+        let exp = expectation(description: "multiple clients should store separately")
```
adamchel: [nit] the expectation description is the same as the previous test

StitchCore/StitchCore/Source/Core/StitchClient.swift
edaniels: I believe there's actually a difference where not specifying `init` causes it do be a coercion that calls init, not directly an init. I'm sure it's compiled to exactly the same thing though.

jsflax: I believe @edaniels is correct. Personally, I do it because I find it to be more legible, and speaks closer to Swift's idioms– verbosity and readability over shortcuts.

StitchCore/StitchCore/Source/Core/StitchClient.swift
```diff
@@ -168,11 +167,32 @@ public class StitchClient: StitchClientType {
...
+                self.storage = MemoryStorage.init(suiteName: suiteName)!
+                printLog(.warning, text: "Invalid suiteName: \(suiteName)")
+                printLog(.warning, text: "Defaulting to memory storage. NOTE: App will not persist data in this state")
```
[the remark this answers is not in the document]
jsflax: Fair point. Will change the wording.

StitchCore/StitchCore/Source/Core/Storage.swift
```diff
@@ -0,0 +1,132 @@
...
+}
+
+/// Failover implementation of the Storage protocol
+/// in the event that no other storage unit can
+/// be found.
+internal struct MemoryStorage: Storage {
+    /// Namespace of this suite
+    private let suiteName: String
+    /// Internal storage unit
+    fileprivate var storage: [String: Any] = [:]
```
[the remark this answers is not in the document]
jsflax: 👍

StitchCore/StitchCoreTests/AuthTests.swift
jsflax: Odd. Alright.

StitchCore/StitchCoreTests/StitchCoreTests.swift
jsflax: Good catch.

StitchCore/StitchCore/Source/Core/Storage.swift
```diff
@@ -0,0 +1,132 @@
...
+        storage[versionKey] = latestVersion
+    #endif
+    // in future versions, `case 1:`, `case 2:` and so on
+    // could be added to perform similar migrations
+    default: break
+    }
+}
+
+internal struct StorageKeys {
+    internal let isLoggedInUDKey: String
```
edaniels: [q] Given the struct is internal, does that not make the internal props internal too? Also, isn't internal the default access anyway?

StitchCore/StitchCore/Source/Core/StitchClient.swift
```diff
@@ -5,41 +5,36 @@ import StitchLogger
 import Security
 import PromiseKit
 
-public struct Consts {
-    public static let DefaultBaseUrl =   "https://stitch.mongodb.com"
-    static let ApiPath =                 "/api/client/v2.0/"
+internal struct Consts {
```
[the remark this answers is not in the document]
jsflax: I believe the original writer did this as a means of namespacing, yes. I've been slowly stripping these constants out, as most of them need to be used dynamically.

I would personally agree and say `internal let`s would be more appropriate here.

StitchCore/StitchCore/Source/Core/Storage.swift
```diff
@@ -0,0 +1,132 @@
...
+    case .none:
+    #if !os(Linux)
+        // map each of the store's keys to a Promise
+        // that fetches the each value for each key,
+        // sets the old value to the new "namespaced" key
+        // remove the old key value pair,
+        // and set the version number
+        // we only care about UserDefaults here since no other
+        // `Storage` type was available before version 1
+        guard var originalStorage = UserDefaults.init(suiteName: "com.mongodb.stitch.sdk.UserDefaults") else {
```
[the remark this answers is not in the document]
jsflax: Technically, it will always "run". But once the value is set, next time this block is hit, the version `case` will be the `default` case, which is empty. If it would be more legible, we could add a simple boolean check to see if `storage.value(forKey: versionKey)` is equal to `latestVersion`.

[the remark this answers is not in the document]
edaniels: oh I missed that. fine as is!

StitchCore/StitchCore/Source/Core/StitchClient.swift
```diff
@@ -5,41 +5,36 @@ import StitchLogger
...
+                       storage: Storage?) -> Promise<TClient>
 }
 
 public final class StitchClientFactory: StitchClientFactoryProtocol {
     typealias TClient = StitchClient
 
     static func create(appId: String,
-                       baseUrl: String = Consts.DefaultBaseUrl,
-                       networkAdapter: NetworkAdapter = StitchNetworkAdapter()) -> Promise<StitchClient> {
+                       baseUrl: String = "https://stitch.mongodb.com",
```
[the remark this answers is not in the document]
jsflax: 👍

StitchCore/StitchCore/Source/Core/Storage.swift
jsflax: This ends up being a style choice.

StitchCore/StitchCore/Source/Core/StitchClient.swift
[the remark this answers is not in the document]
jsflax: I rescind the previous remark, as some of these do need to be exposed via default parameters.

In Swift (I'm not sure why this is the case), if you are using a default parameter, it must be of the same level of access (or higher) than the referenced method.

StitchCore/StitchCore/Source/Core/StitchHTTPClient.swift
```diff
@@ -99,22 +108,23 @@ internal class StitchHTTPClient {
     internal func clearAuth() throws {
         authInfo = nil
 
-        try deleteToken(withKey: Consts.AuthRefreshTokenKey)
-        try deleteToken(withKey: Consts.IsLoggedInUDKey)
-        try deleteToken(withKey: Consts.AuthJwtKey)
+        try deleteToken(withKey: self.storageKeys.authRefreshTokenKey)
+        try deleteToken(withKey: self.storageKeys.isLoggedInUDKey)
```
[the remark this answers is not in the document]
jsflax: I think this should be a separate ticket, but I agree with you.

jsflax:
@adamchel PTAL

PR #30, opened 2018-01-30
STITCH-1040 Improve semantics of multiple logins

adamchel:
In this change I now keep track of the auth provider used to log into a session in a new UserDefaults key, I exposed that in the StitchClient, and I updated `login()` to use the expected semantics.

**Drive-by changes:**
- Updated the enum of AuthProviderTypes to be `internal` rather than `private` so I could compare the types in the `StitchClient`. 
- Changed the hardcoded types in each of the the auth providers to the raw value of the enum type

### DIFF
--- a/StitchCore/StitchCore/Source/Auth/Providers/Anonymous/AnonymousAuthProvider.swift
+++ b/StitchCore/StitchCore/Source/Auth/Providers/Anonymous/AnonymousAuthProvider.swift
@@ -3,7 +3,7 @@ import ExtendedJson
 /// AnonymousAuthProvider provides a way to authenticate anonymously.
 public struct AnonymousAuthProvider: AuthProvider {
     /// The authentication type for anonymous login.
-    public let type: String = AuthProviderTypes.anonymous.rawValue
+    public let type: AuthProviderTypes = AuthProviderTypes.anonymous
     
     /// The JSON payload containing authentication material.
     public var payload: Document = [:]
--- a/StitchCore/StitchCore/Source/Auth/Providers/Anonymous/AnonymousAuthProvider.swift
+++ b/StitchCore/StitchCore/Source/Auth/Providers/Anonymous/AnonymousAuthProvider.swift
@@ -3,7 +3,7 @@ import ExtendedJson
 /// AnonymousAuthProvider provides a way to authenticate anonymously.
 public struct AnonymousAuthProvider: AuthProvider {
     /// The authentication type for anonymous login.
-    public let type: String = "anon-user"
+    public let type: String = AuthProviderTypes.anonymous.rawValue
     
     /// The JSON payload containing authentication material.
     public var payload: Document = [:]
--- a/StitchCore/StitchCore/Source/Auth/Providers/ApiKey/ApiKeyAuthProvider.swift
+++ b/StitchCore/StitchCore/Source/Auth/Providers/ApiKey/ApiKeyAuthProvider.swift
@@ -12,7 +12,7 @@ import ExtendedJson
 public struct ApiKeyAuthProvider: AuthProvider {
     private let key: String
 
-    public var type: String = "api-key"
+    public var type: String = AuthProviderTypes.apiKey.rawValue
     public var name: String = "api-key"
 
     public var payload: Document {
--- a/StitchCore/StitchCore/Source/Auth/Providers/ApiKey/ApiKeyAuthProvider.swift
+++ b/StitchCore/StitchCore/Source/Auth/Providers/ApiKey/ApiKeyAuthProvider.swift
@@ -12,7 +12,7 @@ import ExtendedJson
 public struct ApiKeyAuthProvider: AuthProvider {
     private let key: String
 
-    public var type: String = AuthProviderTypes.apiKey.rawValue
+    public var type: AuthProviderTypes = AuthProviderTypes.apiKey
     public var name: String = "api-key"
 
     public var payload: Document {
--- a/StitchCore/StitchCore/Source/Auth/Providers/Custom/CustomAuthProvider.swift
+++ b/StitchCore/StitchCore/Source/Auth/Providers/Custom/CustomAuthProvider.swift
@@ -4,7 +4,7 @@ import ExtendedJson
 // CustomAuthProvider is a special case that does not
 // follow previous protocols.
 public struct CustomAuthProvider: AuthProvider {
-    public var type: String = AuthProviderTypes.custom.rawValue
+    public var type: AuthProviderTypes = AuthProviderTypes.custom
 
     public var payload: Document { return ["token": jwt] }
 
--- a/StitchCore/StitchCore/Source/Auth/Providers/Custom/CustomAuthProvider.swift
+++ b/StitchCore/StitchCore/Source/Auth/Providers/Custom/CustomAuthProvider.swift
@@ -4,7 +4,7 @@ import ExtendedJson
 // CustomAuthProvider is a special case that does not
 // follow previous protocols.
 public struct CustomAuthProvider: AuthProvider {
-    public var type: String = "custom-token"
+    public var type: String = AuthProviderTypes.custom.rawValue
 
     public var payload: Document { return ["token": jwt] }
 
--- a/StitchCore/StitchCore/Source/Auth/Providers/EmailPassword/EmailPasswordAuthProvider.swift
+++ b/StitchCore/StitchCore/Source/Auth/Providers/EmailPassword/EmailPasswordAuthProvider.swift
@@ -4,9 +4,7 @@ import ExtendedJson
 /// EmailPasswordAuthProvider provides a way to authenticate using an email and password.
 public struct EmailPasswordAuthProvider: AuthProvider {
     /// The authentication type for email/pass login.
-    public var type: String {
-        return AuthProviderTypes.emailPass.rawValue
-    }
+    public var type: AuthProviderTypes = AuthProviderTypes.emailPass
 
     /// The JSON payload containing the username and password.
     public var payload: Document {
--- a/StitchCore/StitchCore/Source/Auth/Providers/EmailPassword/EmailPasswordAuthProvider.swift
+++ b/StitchCore/StitchCore/Source/Auth/Providers/EmailPassword/EmailPasswordAuthProvider.swift
@@ -5,7 +5,7 @@ import ExtendedJson
 public struct EmailPasswordAuthProvider: AuthProvider {
     /// The authentication type for email/pass login.
     public var type: String {
-        return "local-userpass"
+        return AuthProviderTypes.emailPass.rawValue
     }
 
     /// The JSON payload containing the username and password.
--- a/StitchCore/StitchCore/Source/Auth/Providers/Facebook/FacebookAuthProvider.swift
+++ b/StitchCore/StitchCore/Source/Auth/Providers/Facebook/FacebookAuthProvider.swift
@@ -5,7 +5,7 @@ import ExtendedJson
 public struct FacebookAuthProvider: AuthProvider {
     /// The authentication type for facebook login.
     public var type: String {
-        return "oauth2-facebook"
+        return AuthProviderTypes.facebook.rawValue
     }
 
     /// The JSON payload containing the accessToken.
--- a/StitchCore/StitchCore/Source/Auth/Providers/Facebook/FacebookAuthProvider.swift
+++ b/StitchCore/StitchCore/Source/Auth/Providers/Facebook/FacebookAuthProvider.swift
@@ -4,9 +4,7 @@ import ExtendedJson
 /// FacebookAuthProvider provides a way to authenticate via Facebook's OAuth 2.0 provider.
 public struct FacebookAuthProvider: AuthProvider {
     /// The authentication type for facebook login.
-    public var type: String {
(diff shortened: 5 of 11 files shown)


StitchCore/StitchCore/Source/Core/StitchHTTPClient.swift
```diff
@@ -102,6 +106,8 @@ internal class StitchHTTPClient {
         try deleteToken(withKey: Consts.AuthRefreshTokenKey)
         try deleteToken(withKey: Consts.IsLoggedInUDKey)
```
adamchel: I think we should be doing `userDefaults?.removeObject(forKey: Consts.IsLoggedInUDKey)` here because with `deleteToken` it will try to delete the key from the keychain when not using the simulator. Should I make this drive-by change or is this here intentionally?

StitchCore/StitchCore/Source/Core/StitchClient.swift
```diff
@@ -126,6 +129,11 @@ public class StitchClient: StitchClientType {
     public var isAuthenticated: Bool {
         return self.httpClient.isAuthenticated
     }
+    
+    // Returns the type of the provider used to log into the current session. nil if not authenticated
+    public var loggedInProviderType: String? {
+        return self.httpClient.loggedInProviderType
```
edaniels: It seems odd that `loggedInProviderType` is exposed via the HTTP client

StitchCore/StitchCore/Source/Core/StitchClient.swift
```diff
@@ -268,27 +276,41 @@ public class StitchClient: StitchClientType {
     public func login(withProvider provider: AuthProvider) -> Promise<UserId> {
         self.authProvider = provider
 
-        if isAuthenticated, let auth = auth {
-            printLog(.info, text: "Already logged in, using cached token.")
-            return Promise.init(value: auth.userId)
+        func doLoginRequest() -> Promise<UserId> {
+            return httpClient.doRequest {
```
edaniels: [style] This is more for my education, but how much do we personally like the usage of `${0..n}` as opposed to just specifying the parameter in the function closure?

StitchCore/StitchCore/Source/Core/StitchClient.swift
```diff
@@ -268,27 +276,41 @@ public class StitchClient: StitchClientType {
...
-            strongSelf.onLogin()
-            return authInfo.userId
+        if isAuthenticated, let auth = auth {
+            if provider.type == AuthProviderTypes.anonymous.rawValue &&
+                loggedInProviderType == AuthProviderTypes.anonymous.rawValue {
+                printLog(.info, text: "Already logged in as anonymous user, using cached token.")
+                return Promise.init(value: auth.userId)
+            }
+
+            // Using a different provider, log outand then perform login.
```
edaniels: [nit] split `outand`

StitchCore/StitchCore/Source/Core/StitchClient.swift
```diff
@@ -268,27 +276,41 @@ public class StitchClient: StitchClientType {
...
-            guard let strongSelf = self else { throw StitchError.clientReleased }
-            let authInfo = try JSONDecoder().decode(AuthInfo.self,
-                                                    from: JSONSerialization.data(withJSONObject: any))
-            strongSelf.httpClient.authInfo = authInfo
-            strongSelf._auth = Auth(stitchClient: strongSelf,
-                                    stitchHttpClient: strongSelf.httpClient,
-                                    userId: authInfo.userId)
-            strongSelf.onLogin()
-            return authInfo.userId
+        if isAuthenticated, let auth = auth {
```
edaniels: [style] feels odd to scope an `auth` by assigning it from `auth`

StitchCore/StitchCore/Source/Core/StitchClient.swift
```diff
@@ -268,27 +276,41 @@ public class StitchClient: StitchClientType {
...
+            if provider.type == AuthProviderTypes.anonymous.rawValue &&
+                loggedInProviderType == AuthProviderTypes.anonymous.rawValue {
```
edaniels: [note] This will have the effect of logging out an anonymous user since we aren't migrating. This results in a loss of an anonymous account which is okay during our stage of development. We should document it though.

StitchCore/StitchCore/Source/Core/StitchClient.swift
```diff
@@ -268,27 +276,41 @@ public class StitchClient: StitchClientType {
...
+                printLog(.info, text: "Already logged in as anonymous user, using cached token.")
+                return Promise.init(value: auth.userId)
+            }
+
+            // Using a different provider, log outand then perform login.
+            printLog(.info, text: "Already logged in, logging out of existing session.")
+            return self.logout().then {
```
edaniels: [minor] If logout fails, we still want to log in. See how the JS SDK does this in the logout method by handling both resolve and reject cases.

StitchCore/StitchCore/Source/Core/StitchClient.swift
edaniels: [nit] Easier to return early in a small block of code checking `if !isAuthenticated`.

[style?] In fact, it must be the case that auth is not `nil`, which I think means we can use the force unwrap expression `!auth.userId`. @jsflax, as the owners of the state of `auth`, is it fair to say we want a runtime error to happen if `auth` is `nil` when `isAuthenticated` is true or is it too dangerous.

StitchCore/StitchCoreTests/AuthTests.swift
```diff
@@ -115,4 +115,42 @@ class AuthTests: XCTestCase {
 
         wait(for: [exp], timeout: 10)
     }
+    
+    func testMultipleLoginSemantics() throws {
```
edaniels: I suppose there isn't really a good way to test logout failing with our current setup unless a mocking library exists.

StitchCore/StitchCore/Source/Core/StitchClient.swift
adamchel: I did it this way since the HTTP Client is where Auth is actually cleared (and thus the lowest-level part of the client that touches the UserDefaults, but I guess  since nothing in the HTTP Client is accessing `loggedInProviderType` I can move it to `StitchClient`

StitchCore/StitchCore/Source/Core/StitchClient.swift
adamchel: I think the intention here is to ensure that auth isn't nil and thus to avoid doing `auth!`in the if block.

StitchCore/StitchCore/Source/Core/StitchClient.swift
adamchel: Done.

That being said, I'm not sure this is the behavior currently in the JS SDK, because if you look at this line: https://github.com/mongodb/stitch-js-sdk/blob/master/src/client.js#L171

`authenticateFn` is only being called on the resolve case of the `logout` promise. The `logout` method handling both resolve and reject in the JS SDK is for clearing auth in storage, not for authenticating.

StitchCore/StitchCore/Source/Core/StitchClient.swift
adamchel: Would we just put that in release notes somewhere or do you want me to add a comment somewhere?

StitchCore/StitchCore/Source/Core/StitchClient.swift
adamchel: Holding off on @jsflax 's review before making changes here. But if we agree that a runtime error is okay then I can change this to do a `guard isAuthenticated else { return doLoginRequest() }` block instead of the big `if` block.

StitchCore/StitchCore/Source/Core/StitchClient.swift
adamchel: I'm usually for explicitly specifying parameters for the sake of readability but I'll let @jsflax decide since he's more familiar with what's normal in Swift.

StitchCore/StitchCore/Source/Core/StitchClient.swift
edaniels: Release notes

adamchel:
Per discussion with Eric, I made another drive-by change which makes the `logout()` method on the StitchClient semantically identical to `logout()` in the JS SDK. 

Essentially, any error in the logout request—which was previously being thrown as an error to the user—is now being ignored (since the logout request doesn't do anything too important that the user needs to know about). 

The only time an error will be thrown from `logout()` is if there is an issue with clearing the local auth storage, which is really the only important thing `logout()` is doing from an SDK user's perspective.

StitchCore/StitchCore/Source/Auth/Providers/Anonymous/AnonymousAuthProvider.swift
```diff
@@ -3,7 +3,7 @@ import ExtendedJson
 /// AnonymousAuthProvider provides a way to authenticate anonymously.
 public struct AnonymousAuthProvider: AuthProvider {
     /// The authentication type for anonymous login.
-    public let type: String = "anon-user"
+    public let type: String = AuthProviderTypes.anonymous.rawValue
```
jsflax: [minor] All of these can probably be `AuthProviderTypes` now. I'd like to clean this `protocol` up a bit in general, but there's no reason we should be declaring the rawValue for each of these types.

StitchCore/StitchCore/Source/Auth/Providers/ApiKey/ApiKeyAuthProvider.swift
```diff
@@ -12,7 +12,7 @@ import ExtendedJson
 public struct ApiKeyAuthProvider: AuthProvider {
     private let key: String
 
-    public var type: String = "api-key"
+    public var type: String = AuthProviderTypes.apiKey.rawValue
```
jsflax: [minor] Same as above.

StitchCore/StitchCore/Source/Auth/Providers/Custom/CustomAuthProvider.swift
```diff
@@ -4,7 +4,7 @@ import ExtendedJson
 // CustomAuthProvider is a special case that does not
 // follow previous protocols.
 public struct CustomAuthProvider: AuthProvider {
-    public var type: String = "custom-token"
+    public var type: String = AuthProviderTypes.custom.rawValue
```
jsflax: [minor] Same as above.

StitchCore/StitchCore/Source/Auth/Providers/EmailPassword/EmailPasswordAuthProvider.swift
```diff
@@ -5,7 +5,7 @@ import ExtendedJson
 public struct EmailPasswordAuthProvider: AuthProvider {
     /// The authentication type for email/pass login.
     public var type: String {
-        return "local-userpass"
+        return AuthProviderTypes.emailPass.rawValue
```
jsflax: [minor][nit] Same as above, and this doesn't need to be a computed property.

StitchCore/StitchCore/Source/Auth/Providers/Facebook/FacebookAuthProvider.swift
```diff
@@ -5,7 +5,7 @@ import ExtendedJson
 public struct FacebookAuthProvider: AuthProvider {
     /// The authentication type for facebook login.
     public var type: String {
-        return "oauth2-facebook"
+        return AuthProviderTypes.facebook.rawValue
```
jsflax: [minor][nit] Same as above, and this doesn't need to be a computed property.

StitchCore/StitchCore/Source/Auth/Providers/Google/GoogleAuthProvider.swift
```diff
@@ -5,7 +5,7 @@ import ExtendedJson
 public struct GoogleAuthProvider: AuthProvider {
     /// The authentication type for google login.
     public var type: String {
-        return "oauth2-google"
+        return AuthProviderTypes.google.rawValue
```
jsflax: [minor][nit] Same as above, and this doesn't need to be a computed property.

StitchCore/StitchCore/Source/Core/StitchClient.swift
jsflax: Conditional bindings also allow inner properties of structures, so we could just strip out `userId`. Adam's idea of using a `guard` here also works.

StitchCore/StitchCoreTests/AuthTests.swift
jsflax: Correct ... I may just push the Admin client forward sooner than later at this point. These tests are only going to be flimsy until then.

This change will come alongside a mocking framework (which we'll have to pick) and a matcher framework (https://github.com/Quick/Nimble) which will make our testing much more robust.

adamchel:
Just went through Jason’s comments. In addition to the requested changes:

- `AuthProviderTypes` is now a `public` enum (was originally `private`, then `internal` for my first pass)
- I added an informational log message that happens when the request  in `logout()` fails.
- The computed `loggedInProviderType` property on `StitchClient` now returns an `Optional(AuthProviderTypes)` rather than an `Optional(String)`

PR #31, opened 2018-01-31
STITCH-1140 iOS SDK: Create factory for clients

jsflax:


jsflax:
Yeah I think I'll change it then.

PR #32, opened 2018-02-01
STITCH-1075: iOS SDK Add ability for a user to link to any non-anonymous identity

adamchel:
This change adds linking functionality to the iOS SDK. Should be functionally identical to the JS SDK changes, except that linking in the iOS SDK does not attempt to update the underlying auth object of StitchClient. I also made `loggedInProvider` always reflect the most recently linked identity to match the behavior of the JS SDK.

I tested this locally with an emal/password identity but until we can use the admin API in our tests it’s going to be difficult to test this. I left the test in as a comment.

**Drive-bys:**
- Made StitchClient in multiple login semantics test conform to new factory pattern
- Made `deviceID` and `refreshToken` optional in `AuthInfo` struct so the `decode()` doesn’t fail when getting the linking auth response
- refactored the `doLoginRequest()` closure from `login()` into a `doAuthRequest()` method that is called from both `login()` and `link()` with slightly different parameters.

### DIFF
--- a/StitchCore/StitchCore/Source/Auth/AuthInfo.swift
+++ b/StitchCore/StitchCore/Source/Auth/AuthInfo.swift
@@ -1,7 +1,19 @@
 import Foundation
 
+protocol AuthResponse: Codable {
+    var userId: String { get }
+}
+
+internal struct LinkInfo: AuthResponse {
+    enum CodingKeys: String, CodingKey {
+        case userId = "user_id"
+    }
+    
+    var userId: String
+}
+
 /// Auth represents the current authorization state of the client
-internal struct AuthInfo: Codable {
+internal struct AuthInfo: AuthResponse {
     enum CodingKeys: String, CodingKey {
         case accessToken = "access_token",
         userId = "user_id",
@@ -14,13 +26,13 @@ internal struct AuthInfo: Codable {
     let accessToken: DecodedJWT?
 
     // The user this session was created for.
-    let deviceId: String?
+    let deviceId: String
 
     // The user this session was created for.
     let userId: String
 
     // The refresh token to refresh an expired access token
-    internal var refreshToken: String?
+    internal var refreshToken: String
 
     internal func auth(with updatedAccessToken: String) -> AuthInfo {
         return AuthInfo(accessToken: try? DecodedJWT(jwt: updatedAccessToken),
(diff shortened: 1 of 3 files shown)


StitchCore/StitchCore/Source/Core/StitchClient.swift
```diff
@@ -367,10 +346,61 @@ public class StitchClient: StitchClientType {
...
+                               withLinking linking: Bool = false) -> Promise<UserId> {
+        return httpClient.doRequest { request in
+            request.method = .post
+
+            let authRoute = self.routes.authProvidersLoginRoute(provider: provider.type.rawValue)
+            request.endpoint = "\(authRoute)\(linking ? "?link=true" : "")"
+            request.isAuthenticatedRequest = linking
+
+            try request.encode(withData: self.getAuthRequest(provider: provider))
+            }.flatMap { [weak self] any in
```
jsflax: [nit] Indentation

jsflax: [nit] Let's call `any` `json`

StitchCore/StitchCoreTests/AuthTests.swift
```diff
@@ -187,4 +191,41 @@ class AuthTests: XCTestCase {
         
         wait(for: [exp], timeout: 10)
     }
+
+    // NOTE: This test works locally when logging in with an email identity not associated with a Stitch user, but with our
+    // current testing framework we cannot dynamically create identities to test this functionality. Once we have the
+    // appropriate framework we can re-enable this test.
+    /*
```
jsflax: [q] What happened here?

StitchCore/StitchCore/Source/Auth/AuthInfo.swift
```diff
@@ -14,13 +14,13 @@ internal struct AuthInfo: Codable {
     let accessToken: DecodedJWT?
 
     // The user this session was created for.
-    let deviceId: String
+    let deviceId: String?
```
jsflax: [q] What happened here?

StitchCore/StitchCore/Source/Auth/AuthInfo.swift
```diff
@@ -14,13 +14,13 @@ internal struct AuthInfo: Codable {
...
 
     // The user this session was created for.
     let userId: String
 
     // The refresh token to refresh an expired access token
-    internal var refreshToken: String
+    internal var refreshToken: String?
```
jsflax: [q] What happened here?

StitchCore/StitchCore/Source/Auth/AuthInfo.swift
jsflax: An offline conversation will be had to reject DRY principles here in favor of creating a separate struct for linking.

adamchel:
@jsflax PTAL

PR #36, opened 2018-02-06
STITCH-1176 iOS SDK: Create Admin Client

jsflax:


StitchCore/StitchCoreTests/Admin/StitchAdminClient.swift
```diff
@@ -0,0 +1,186 @@
...
+extension Remove {
+    public func remove() -> Promise<Any> {
+        return self.httpClient.doRequest {
+            $0.endpoint = self.url
+            $0.method = .delete
+        }
+    }
+}
+
+// Adds an endpoint method that POSTs new data
```
edaniels: [harsh-opinion] I'm not convinced that HTTP methods and actions should map to protocols. They are saving code, but that could still easily be done with a more succinct doRequest interface. I feel like this adds needless complexity and is slightly an abuse of protocols despite it looking cool.

This opinion applies to `Endpoint`, `List`, `Get`, `Remove`, `Create`, `Update`, `Enable`, and `Disable`

StitchCore/StitchCoreTests/Admin/StitchAdminClient.swift
```diff
@@ -0,0 +1,186 @@
+import Foundation
+import PromiseKit
+import ExtendedJson
+@testable import StitchCore
+
+// Any endpoint that can be described with basic
+// CRUD operations
+internal protocol Endpoint {
```
edaniels: I'm slightly okay with this protocol if all it was was a url property. `httpClient` seems like an odd coupling for describing something that conforms to an "endpoint"

StitchCore/StitchCoreTests/Admin/StitchAdminClient.swift
```diff
@@ -0,0 +1,186 @@
...
+}
+
+public class StitchAdminClient {
+    let baseUrl: String
+    let httpClient: StitchHTTPClient
+
+    fileprivate init(baseUrl: String) {
+        self.baseUrl = baseUrl
+        self.httpClient = StitchHTTPClient.init(baseUrl: baseUrl,
+                                                apiPath: "/api/admin/v3.0",
```
edaniels: [nit] refactor out into a const

StitchCore/StitchCoreTests/Admin/StitchAdminClient.swift
```diff
@@ -0,0 +1,186 @@
...
+                                                networkAdapter: StitchNetworkAdapter(),
+                                                storage: MemoryStorage(),
+                                                storageKeys: StorageKeys.init(suiteName: "__admin__"))
+    }
+
+    func apps(withGroupId groupId: String) -> AppsEndpoint {
```
edaniels: [opinion] From an API perspective, receiving an "endpoint" doesn't feel right. Why don't we break from what the JS SDK did and not offer a hierarchy of "endpoints"? I think it makes the world confusing versus just `adminClient.getApp(:withAppId:inGroupId)` or something. That would significantly decrease the complexity and remove the need for endpoints and many protocols as well as decouple httpclients from endpoints. You could build each of these methods as grouped extensions in a different file.

StitchCore/StitchCoreTests/Admin/Services/ServicesEndpoint.swift
```diff
@@ -0,0 +1,42 @@
+import Foundation
```
edaniels: [harsh-opinion] ibid

StitchCore/StitchCoreTests/Admin/RulesEndpoint.swift
```diff
@@ -0,0 +1,84 @@
+import Foundation
```
edaniels: [harsh-opinion] ibid

StitchCore/StitchCoreTests/Admin/AuthProviders/AuthProvidersEndpoints.swift
```diff
@@ -0,0 +1,55 @@
+import Foundation
```
edaniels: [harsh-opinion] ibid

StitchCore/StitchCoreTests/Admin/UserRegistrationsEndpoints.swift
```diff
@@ -0,0 +1,33 @@
+import Foundation
```
edaniels: [harsh-opinion] ibid

StitchCore/StitchCoreTests/Admin/UsersEndpoints.swift
```diff
@@ -0,0 +1,42 @@
+import Foundation
```
edaniels: [harsh-opinion] ibid

StitchCore/StitchCoreTests/AuthTests.swift
```diff
@@ -1,97 +1,48 @@
...
-            exp.fulfill()
-        }.catch { err in
-            print(err)
-            XCTFail(err.localizedDescription)
-            exit(1)
-        }
-
-        wait(for: [exp], timeout: 200)
+        try await(stitchClient.login(withProvider: AnonymousAuthProvider()))
+        let userProfile = try await(self.stitchClient.auth!.fetchUserProfile())
```
edaniels: [q] From a testing perspective, will a force unwrap on an `nil` `Optional` end the whole test process or just this test? If the former, we should probably address.

StitchCore/StitchCoreTests/StitchTestCase.swift
```diff
@@ -0,0 +1,73 @@
...
+                                     tokenId: conf.tokenId)
+        }.asVoid()
+    }
+}
+
+enum StitchTestError: Error {
+    case unknown
+}
+
+open class StitchTestCase: XCTestCase {
```
edaniels: [remark] Would be cool if you were allowed to be more restrictive here (i.e. make this internal)

StitchCore/StitchCoreTests/TestHarness/TestHarness.swift
```diff
@@ -0,0 +1,172 @@
...
+//  Created by Jason Flax on 2/1/18.
+//  Copyright © 2018 MongoDB. All rights reserved.
+//
+
+import Foundation
+@testable import StitchCore
+import PromiseKit
+import ExtendedJson
+
+internal let defaultServerUrl = "http://localhost:9090"
```
edaniels: We can do this later but this is going to want to default to prod or be lazily initialized based on environment settings. Same goes for username and password. For normal users, they can't even do username and password but only "api-key".

StitchCore/StitchCoreTests/TestHarness/TestHarness.swift
```diff
@@ -0,0 +1,172 @@
...
+                return self.app.services.service(withId: view.id).rules.create(data: $0)
+            }).flatMap { _ in view }
+        }
+    }
+    
+    func add(providerConfig: ProviderConfigs) -> Promise<AuthProviderView> {
+        return self.app.authProviders.create(data: providerConfig)
+    }
+
+    func addDefaultUserpassConfig() -> Promise<AuthProviderView> {
```
edaniels: [nit] We're not adding a config but adding a provider (we just provide a config).

StitchCore/StitchCoreTests/Admin/AuthProviders/AuthProvidersEndpoints.swift
```diff
@@ -0,0 +1,55 @@
...
+@testable import StitchCore
+
+public struct AuthProviderView: Codable {
```
adamchel: [style] ditto about `View` suffix

StitchCore/StitchCoreTests/Admin/RulesEndpoint.swift
```diff
@@ -0,0 +1,84 @@
...
+                                     head: head).encode(to: encoder)
+        case .twilio(let send):
+            try TwilioRuleActions.init(send: send).encode(to: encoder)
+        case .awsSes(let send):
+            try AwsSesRuleActions.init(send: send).encode(to: encoder)
+        }
+    }
+}
+
+internal struct RuleView: Codable {
```
adamchel: [style] ditto about `View` suffix

[q] also is this struct supposed to be empty? Seems like Get and List would end up being be empty for `RuleEndpoint` and `RulesEndpoint`.

StitchCore/StitchCoreTests/Admin/Services/ServicesEndpoint.swift
```diff
@@ -0,0 +1,42 @@
...
+@testable import StitchCore
+
+internal struct ServiceView: Codable {
```
adamchel: [style] ditto about `View` suffix

StitchCore/StitchCoreTests/Admin/UsersEndpoints.swift
```diff
@@ -0,0 +1,42 @@
...
+@testable import StitchCore
+
+struct UserCreator: Encodable {
+    let email: String
+    let password: String
+}
+struct UserView: Decodable {
```
adamchel: [style] Ditto about `View` suffix

StitchCore/StitchCoreTests/TestHarness/TestHarness.swift
```diff
@@ -0,0 +1,172 @@
...
+
+    func addDefaultAnonConfig() -> Promise<AuthProviderView> {
+        return self.add(providerConfig: .anon())
+    }
+
+    func addDefaultCustomTokenConfig() -> Promise<AuthProviderView> {
+        return self.add(providerConfig: .custom(signingKey: "abcdefghijklmnopqrstuvwxyz1234567890"))
+    }
+
+    func setupStitchClient(shouldConfigureUserAuth: Bool = true) -> Promise<Void> {
```
adamchel: [nit] param unnecessary or do we still want to implement the setup of an unauthed StitchClient?

adamchel: [q] on a similar note, shouldn't our linter or the swift compiler be catching these?

StitchCore/StitchCore/Source/Auth/UserProfile.swift
```diff
@@ -33,4 +42,5 @@ public struct UserProfile: Codable {
     public let identities: [Identity]
     /// The extra data associated with this user.
     public let data: [String: String]
+    internal let roles: [Role]?
```
adamchel: I don't know how I feel about making `UserProfile` be the model for both regular Stitch users and Admin users, esp. since this property will be irrelevant for 99% of our users (unless we're adding roles to regular Stitch users too). Should we reject DRY principles here like we did for `LinkInfo` vs. `AuthInfo` and make an `AdminUserProfile` class in `StitchCoreTests`?

edaniels: It is internal though so no external user can use it

StitchCore/StitchCoreTests/Admin/AppEndpoint.swift
```diff
@@ -0,0 +1,69 @@
+import Foundation
+import PromiseKit
+@testable import StitchCore
+
+struct AppView: Codable {
```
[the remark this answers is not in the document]
edaniels: Given this is in a different module, I'm okay with `App`. If it were in StitchCore, I think it would be too confusing. To play devil's advocate against that though, I wouldn't put it past a user to think `App` has methods on it like delete or addService whereas `AppModel` would work.

Fun hack and possible abuse idea:
`struct Model {}`
`extension Model { struct App {} }`

:)

adamchel:
So after some more thought I think I’m actually in favor of Jason’s architecture, with some significant changes.

I tried imagining what the doRequest architecture would look like and it’s not much prettier (though @edaniels please jump in if you had something different in mind). First I imagine we’d have the following methods either on the admin client or http client:

```
doGetRequest //(generic with Model associated type)
doGetListRequest //(generic with Model associated type)
doPostRequest //(generic with CreatorModel and Model associated types)
doPutRequest //(generic with Model associated type)
doDeleteRequest
doEnableRequest
doDisableRequest
```

Then in files where you have grouped extension functions you’d probably have something like:
```
AppRequests.swift
	// extensions of StitchAdminClient
	createApp() // calls custom request code
	listApps() // calls doGetListRequest<[AppModel]>()
	getApp(appID)
	removeApp(appId)
	
UserRequests.swift:
	// extensions of StitchAdminClient
	getUser(appId, userId) // calls doGetRequest<UserModel>()
	removeUser(appId, userId) // calls doDeleteRequest()
	listUsers(appId) // calls doGetListRequest<[UserModel]>()
	createUser(appId) // calls doPostRequest <UserCreator, UserModel>()
	sendUserConfirmation(appId) // calls custom request code

ServiceRequests.swift:
	// extensions of StitchAdminClient
	getService(appId, serviceId) // calls doGetRequest<ServiceModel>
	removeService(appId) // you get the idea…
	listServices(appId)
	createService(appId)

RuleRequests.swift
	getRule(appId, serviceId, ruleId)
	removeRule(appId, serviceId, ruleId)
	listRules(appId, serviceId)
	createRule(appId, serviceId)
```

This gets a little messy, esp with rules since we’re taking so many IDs.

The main advantage of Jason’s model is that it keeps the StitchAdminClient clean and hierarchical, so even if we end up having hundreds of possible actions in the admin API with deeply nested actions it still mostly makes sense and we don’t have to keep track of all the IDs.

If we’re gonna go with the endpoint protocol solution though, I think we should do the following:

- Add “-able” suffixes to the protocols. I think this would make the fact code more clear and seems to fit with the Swift naming conventions.
```
Gettable
Listable
Createable
Updateable
Removable
Enableable
Disableable
```

- Change “Endpoint” suffixes to “Resource” or “Source”. Reasoning for this is you don’t “get” or “update” an “endpoint”, you “get” or “update” a “resource”. With the “resource” name, the coupling of an HTTP client and URL makes a bit more sense.

```
AppResource
AuthProviderResource
AuthProvidersResource
UserResource
UsersResource
//etc…
```

- Let’s make something like a `BasicEndpoint` class that implements `Endpoint` (`BasicResource` implementing `Resource` if we’re changing the names), so that each of the AppEndpoint/AppResource classes can just be a subclass of the Basic Endpoint so they don’t each need to manage their own HTTPClient. 

```
public class BasicResource: Resource {
    internal let httpClient: StitchHTTPClient
    internal let url: String
    
    internal init(httpClient: StitchHTTPClient,
                     url: String) {
        self.httpClient = httpClient
        self.url = url
    }
}
```

This way, AppResource can just look something like this:

```swift
public final class AppResource: BasicResource, Gettable, Removeable {
    typealias Model = AppView

    lazy var authProviders: AuthProvidersResource =
        AuthProvidersEndpoint.init(httpClient: self.httpClient,
                                   url: "\(self.url)/auth_providers")

    lazy var users: UsersResource =
        UsersEndpoint.init(httpClient: self.httpClient,
                           url: "\(self.url)/users")

    lazy var userRegistrations: UserRegistrationsResource =
        UserRegistrationsResource.init(httpClient: self.httpClient,
                                       url: "\(self.url)/user_registrations")

    lazy var services: ServicesResource =
        ServicesResource.init(httpClient: self.httpClient, url: "\(self.url)/services")
}
```

and less complex resources could be as simple as this:

```swift
public final class UserResource: BasicResource, Gettable, Removeable {
    typealias Model = UserView
}

public final class UsersResource: BasicResource, Listable, Createable {
    typealias Model = UserView
    typealias CreatorModel = UserCreator

    func user(uid: String) -> UserResource {
        return UserResource.init(httpClient: self.httpClient,
                                 url: "\(url)/\(uid)")
    }
}
```
And this could be even better with a `fromParent` constructor on the BasicEndpoint/BasicResource:
```swift
public class BasicResource: Resource {
    internal let httpClient: StitchHTTPClient
    internal let url: String
    
    internal init(httpClient: StitchHTTPClient,
                  url: String) {
        self.httpClient = httpClient
        self.url = url
    }
    
    internal init(fromParent parent: BasicResource, url: String) {
        self.httpClient = parent.httpClient
        self.url = "\(parent.url)/\(url)"
    }
}

public final class UsersResource: BasicResource, Listable, Createable {
    typealias Model = UserView
    typealias CreatorModel = UserCreator

    func user(uid: String) -> UserEndpoint {
        return UserEndpoint.init(fromParent: self,
                                 url: "/\(uid)")
    }
}
```

Also this is all just on my limited Swift knowledge as well, so I’m very open to other suggestions/rebuttals. If anything doesn’t make sense lmk and I can try to clarify.

.gitmodules
```diff
@@ -4,3 +4,4 @@
 [submodule "StitchCore/PromiseKit"]
 	path = StitchCore/PromiseKit
 	url = https://github.com/mxcl/PromiseKit/
+
```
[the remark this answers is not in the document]
jsflax: 👍

StitchCore/StitchCore/Source/Core/StitchClient.swift
```diff
@@ -17,7 +17,7 @@ public struct Consts {
     static let ErrorKey = "error"
 }
 
-internal protocol StitchClientFactoryProtocol {
+protocol StitchClientFactoryProtocol {
```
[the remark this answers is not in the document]
jsflax: Artifact leftover from some in-progress changes. Going to change back.

StitchCore/StitchCore/Source/Core/StitchClient.swift
```diff
@@ -320,6 +320,27 @@ public class StitchClient: StitchClientType {
      */
     @discardableResult
     public func login(withProvider provider: AuthProvider) -> Promise<UserId> {
+        self.authProvider = provider
```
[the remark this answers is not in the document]
jsflax: This seems to be part of some weird merge conflict. Will sort this out.

StitchCore/StitchCore/Source/Core/StitchHTTPClient.swift
```diff
@@ -75,7 +76,6 @@ internal class StitchHTTPClient {
 
     internal func save(token: String, withKey key: String) {
         if isSimulator {
-            printLog(.debug, text: "Falling back to saving token in UserDefaults because of simulator bug")
```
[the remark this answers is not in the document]
jsflax: When testing, this is printed constantly. Considering that we're the only ones running on a simulator, it would be nice to remove this clutter.

StitchCore/StitchCoreTests/Admin/AppEndpoint.swift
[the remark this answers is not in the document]
jsflax: 👍

StitchCore/StitchCoreTests/Admin/StitchAdminClient.swift
jsflax: 👍

StitchCore/StitchCore/Source/Core/Storage.swift
```diff
@@ -12,30 +12,30 @@ private let versionKey = "__stitch_storage_version__"
 internal func runMigration(suiteName: String, storage: inout Storage) {
```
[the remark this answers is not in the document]
jsflax: 👍

StitchCore/StitchCoreTests/AuthTests.swift
jsflax: I believe it will end the test. I guess in theory, you should never have that crash.

StitchCore/StitchCoreTests/StitchTestCase.swift
jsflax: 👍

StitchCore/StitchCoreTests/TestHarness/TestHarness.swift
```diff
@@ -0,0 +1,172 @@
+//
+//  TestHarness.swift
+//  StitchCoreTests
+//
+//  Created by Jason Flax on 2/1/18.
```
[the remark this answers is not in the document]
jsflax: 👍

StitchCore/StitchCoreTests/TestHarness/TestHarness.swift
jsflax: 👍

StitchCore/StitchCoreTests/TestHarness/TestHarness.swift
jsflax: 👍

StitchCore/StitchCoreTests/TestHarness/TestHarness.swift
jsflax: Our linter isn't set up for the testing traget. That's a discussion we can have at the architecture meeting.

jsflax:
@edaniels @adamchel PTAL

StitchCore/StitchCore/Source/Core/StitchHTTPClient.swift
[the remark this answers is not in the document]
adamchel: [nit] We should at least keep it as a comment so there's some context as to why the logic is there.

StitchCore/StitchCoreTests/Admin/Resources.swift
```diff
@@ -0,0 +1,193 @@
...
+    public func disable() -> Promise<Any> {
+        return self.httpClient.doRequest {
+            $0.endpoint = "\(self.url)/disable"
+            $0.method = .put
+        }
+    }
+}
+
+/// Resource that lists the current groupId's applications
+internal final class AppsResource: BasicResource, Listable {
```
adamchel: [style q] This structure is nice and a lot more understandable, but for the sake of making it cleaner should we just drop the "Resource" suffix so referring to the Rule class here is `Apps.App.Services.Rules.Rule`? The suffix seems redundant now that everything inherits `BasicResource and it might make things elsewhere look cleaner.

StitchCore/StitchCoreTests/Admin/Apps/AppsResources.swift
```diff
@@ -0,0 +1,39 @@
+import Foundation
+import PromiseKit
+@testable import StitchCore
+
+/// View into a specific application
+struct AppView: Codable {
```
adamchel: [q] Any reason we stayed with the `View` suffix and not something else?

PR #42, opened 2018-03-28
STITCH-1347 add SwiftLint, and update .gitignore

adamchel:
Added SwiftLint build phases to StitchCore, StitchCore-iOS, and StitchSDKExample, made everything conform, and added exceptions for certain variable names, and for force tries in the case of the `sync` function.

**Drive-by Fixes:**
- updated the .gitignore to the one we were using in the proto repo


adamchel:
The other PR is rebased on this one, so I'm going to merge as is, and then resolve all the comments in the other PR.

PR #43, opened 2018-03-30
STITCH-1294 v4 iOS SDK Documentation

adamchel:
This is most of the docs work that needs to be done for iOS. Remaining work will be distribution (STITCH-1353), and renaming modules and documenting how to get the package in XCode once we work on packaging (STITCH-1293). 

To generate the documentation for a particular module, run `jazzy -c —config jazzy.json` in the root directory of the module. Right now both `StitchCore` and `StitchCore-iOS` are documented and have jazzy configurations.

For public facing protocols/classes/etc, I tried to be as comprehensive as possible with documentation markup.

For internal protocols/classes/etc, I tried to explain every symbol, but I did not necessarily use the documentation markup to describe every parameter, return value, or thrown exception, because in many cases it was self-explanatory.

The only significant thing left undocumented is a richer description of each error case in `StitchErrorCode`. I’m happy to work on it, jbut didn’t think it was worth the time right now.

Also, due to a bug in Jazzy (https://github.com/realm/jazzy/issues/896), I had to put all public-facing function signatures in protocols on one line and add SwiftLint exceptions where that made the line more than 120 characters long. Until the Jazzy bug is fixed, we’ll need to maintain those linter exceptions.

I also made a number of drive-by changes that were mostly refactoring/cleanup. They’re all described here:

**Major drive-by code changes:**
- Moved refresh token related methods in `CoreStitchAuth` to CoreStitchAuth+StitchAuthRequestClient.swift
- Moved the `shouldRefreshOnFailure` property to `StitchAuthRequest` from `StitchRequest` since the property is only relevant for authenticated requests
- Removed `StitchRequestBuilderError` since everything was just using `StitchRequestBuilder` anyway.
- Removed redundant `url` property in StitchRequest and StitchAuthRequest
- Renamed `StitchAuthDocRequestBuilderError` to `StitchDocRequestBuilderError` and moved it to `StitchRequest.swift` since it isn’t exclusive to authenticated requests.
- Renamed `StitchAuthRoutes.swift` to `StitchAppRoutes.swift` since they weren’t exclusively about auth. 
- Moved builder and Buildee protocols to their own file in `Common/Builder.swift` since they’re used by both requests and client configurations
- Made `StitchErrorCodable` internal rather than public.
- Made `StitchUserProfileImpl` internal rather than public.
- Renamed `StitchClientErrors` to `StitchClientError`
- Made `StitchCore-iOS` a shared scheme in the workspace so jazzy could build it properly from the command line with `xcodebuild`.
- Changed the “args” field passed to the Stitch server in requests to “arguments”


StitchCore-iOS/DocsSources/Abstracts/Getting Started.md
```diff
@@ -0,0 +1,82 @@
...
+4. Once the app is running, open up the Debug Area by going to View, Debug Area, Show Debug Area.
+5. You should see a log message like:
+
+    ```
+    Logged in as user 58c5d6ebb9ede022a3d75050
+    ```
+
+#### Executing a Function
+
+1. Once logged in, executing a function happens via the StitchClient's `executeFunction()` method
```
edaniels: [typo] `executeFunction`

StitchCore-iOS/StitchCore-iOS/Core/Auth/Internal/StitchAuthImpl.swift
```diff
@@ -26,78 +52,172 @@ internal final class StitchAuthImpl: CoreStitchAuth<StitchUserImpl>, StitchAuth
...
+     *
+     * - parameters:
+     *     - forProvider: The authentication provider conforming to `AuthProviderClientSupplier` which will provide the
+     *                    client for this authentication provider. Use the `clientProvider` field of the desired
+     *                    authentication provider class.
+     * - returns: an authentication provider client whose type is determined by the `Client` typealias in the type
+     *            specified in the `forProvider` parameter.
+     */
+    public func providerClient<Provider>(forProvider provider: Provider)
+        -> Provider.Client where Provider: AuthProviderClientSupplier {
```
edaniels: [q] It's called a `AuthProviderClientSupplier` but the instances are a `clientProvider`. Should we change one or the other?

StitchCore-iOS/StitchCore-iOS/Core/Auth/Internal/StitchUserFactoryImpl.swift
```diff
@@ -1,7 +1,14 @@
 import Foundation
 import StitchCore
 
+/**
+ * An implementation of `StitchCore.StitchUserFactory`, capable of producing `StitchUserImpl` objects.
+ */
 internal final class StitchUserFactoryImpl: StitchUserFactory {
+    /**
+     * The factory function which can produce a `StitchUserImpl` with the provided ID, logged in provider type/name,
```
edaniels: [nit] ID -> id.

StitchCore/Tests/stitch-ios-sdk-protoTests/Internal/Common/SynchronizedTests.swift
```diff
@@ -11,21 +12,25 @@ class SynchronizedTests: XCTestCase {
...
 
         let sema = DispatchSemaphore(value: 0)
-        Thread {
-            for i in 0 ..< 100 {
-                queue.addOperation {
-                    try! sync(self) {
-                        let value = counter.value
-                        if i % 2 == 0 {
-                            usleep(50)
+        if #available(OSX 10.12, *) {
```
edaniels: [q] Why was this added?

StitchCore/Sources/StitchCore/StitchClientConfiguration.swift
```diff
@@ -1,17 +1,54 @@
...
+ * Properties representing the configuration of a client that can communicate with MongoDB Stitch.
+ */
 public protocol StitchClientConfiguration {
+    /**
+     * The base URL of the Stitch server that the client will communicate with.
+     */
     var baseURL: String { get }
+
+    /**
+     * The underlying storage for authentication info.
```
edaniels: [nit] It's kind of more than just for auth info. We store device details too. Maybe `The underlying storage for persisting authentication and app state.`

StitchCore/Sources/StitchCore/StitchAppClientInfo.swift
```diff
@@ -1,16 +1,37 @@
...
     }
-    
+
+    /**
+     * The client app ID of the Stitch application that this client communicates with.
+     */
     public let clientAppId: String
+
+    /**
+     * The local data directory for any attached instance of mobile MongoDB.
```
edaniels: [nit] This is more generally the data directory Stitch is allowed to use to store any data. Mobile MongoDB will get its data stored here however.

StitchCore/Sources/StitchCore/StitchAppClientInfo.swift
```diff
@@ -1,16 +1,37 @@
...
+                localAppVersion: String) {
         self.clientAppId = clientAppId
         self.dataDirectory = dataDirectory
         self.localAppName = localAppName
         self.localAppVersion = localAppVersion
     }
-    
+
+    /**
+     * The client app ID of the Stitch application that this client communicates with.
```
edaniels: [nit] All IDs -> to ids

StitchCore/Sources/StitchCore/Auth/Providers/UserPass/CoreUserPasswordAuthProviderClient.swift
```diff
@@ -1,25 +1,57 @@
...
         self.authRoutes = authRoutes
         self.providerName = providerName
     }
 
-    private func extensionRoute(forPath path: String) -> String{
+    /**
+     * Private helper which returns the provided path, appended to the base route for the authentication provider.
+     */
+    private func extensionRoute(forPath path: String) -> String {
         return "\(authRoutes.authProviderLoginRoute(withProviderName: providerName))/\(path)"
```
edaniels: [minor] Not code you wrote but I'm pretty sure this has to be wrong. The extension comes after the auth provider name, not `/login`.

edaniels: @jsflax, this is where the spec would come into handy if it defined literally the URL format that must match.

StitchCore-iOS/StitchCore-iOS/Core/Stitch.swift
```diff
@@ -33,67 +39,26 @@ public class Stitch {
     public static func initialize() throws {
         try! StitchCore.sync(self) {
```
edaniels: [aside] @jsflax do we have a ticket for potentially eliminating these `try!`s?

StitchCore-iOS/StitchCore-iOS/Services/Internal/NamedServiceClientProvider.swift
```diff
@@ -1,19 +1,39 @@
...
+    /**
+     * Returns a client of type `ClientType`, with the provided `StitchService` and `StitchAppClientInfo` objects.
+     */
+    func client(forService service: StitchService, withClient client: StitchAppClientInfo) -> ClientType
 }
 
+/**
+ * A generic wrapper for a `NamedServiceClientProvider`.
+ */
 public struct AnyNamedServiceClientProvider<T> {
```
edaniels: [q] Why is this different than how the auth suppliers work?

StitchCore-iOS/StitchCore-iOS/Services/Internal/StitchServiceImpl.swift
```diff
@@ -13,6 +24,19 @@ internal final class StitchServiceImpl: CoreStitchService, StitchService {
...
+     *     - withArgs: The `BSONArray` of arguments to be provided to the function.
+     *     - completionHandler: The completion handler to call when the function call is complete.
+     *                          This handler is executed on a non-main global `DispatchQueue`.
+     *     - result: The result of the function call as an `Any`, or `nil` if the function call failed.
+     *     - error: An error object that indicates why the function call failed, or `nil` if the function call was
+     *              successful.
+     *
+     */
     public func callFunction(withName name: String,
                              withArgs args: BSONArray,
```
edaniels: [aside] Won't this be onerous for callers to construct arrays rather than a variadic list of some sort?

StitchCore/Common/Builder.swift
```diff
@@ -0,0 +1,34 @@
+/**
```
edaniels: [q] How useful is this abstraction?

StitchCore/Sources/StitchCore/Auth/Internal/Models/StoreAuthInfo.swift
```diff
@@ -1,26 +1,53 @@
...
+     */
     let deviceId: String
 
+    /**
+     * The temporary access token for the user.
+     */
     let accessToken: String
 
+    /**
+     * The permanent refresh token for the user.
```
edaniels: [nit] A refresh token has the potential to be invalidated.

StitchCore/Sources/StitchCore/Auth/Providers/Anonymous/AnonymousCredential.swift
```diff
@@ -1,20 +1,34 @@
...
 public struct AnonymousCredential: StitchCredential {
+    /**
+     * The name of the provider for this credential.
+     */
     public var providerName: String
 
+    /**
+     * The type of the provider for this credential.
+     */
     public var providerType: String = "anon-user"
```
edaniels: [nit] All of these `providerType` should be constants.

StitchCore-iOS/StitchCore-iOS/Core/Auth/Internal/StitchAuthImpl.swift
```diff
@@ -2,19 +2,45 @@ import ExtendedJSON
...
+     */
     private struct DelegateWeakRef {
         weak var value: StitchAuthDelegate?
         init(value: StitchAuthDelegate) {
             self.value = value
         }
     }
+
+    /**
+     * An array of weak references to `StitchAuthDelegate`, each of which will be notified when authentication events
```
jsflax: [nit] list* not array

StitchCore-iOS/StitchCore-iOS/Core/Auth/Providers/UserPassword/UserPasswordAuthProvider.swift
```diff
@@ -88,45 +99,54 @@ public protocol UserPasswordAuthProviderClient {
...
     }
-    
-    func confirmUser(withToken token: String, withTokenId tokenId: String, completionHandler: @escaping (Error?) -> Void) {
+
+    func confirmUser(withToken token: String,
+                     withTokenId tokenId: String,
+                     completionHandler: @escaping (Error?) -> Void) {
         dispatcher.run(withCompletionHandler: completionHandler) {
-            let _ = try super.confirmUser(withToken: token, withTokenId: tokenId)
+            _ = try super.confirmUser(withToken: token, withTokenId: tokenId)
```
jsflax: [nit][q] Do we need the `_ =` syntax here? Without a return value, there shouldn't be a warning here.

StitchCore-iOS/StitchCore-iOS/Core/Auth/Providers/UserPassword/UserPasswordAuthProvider.swift
```diff
@@ -88,45 +99,54 @@ public protocol UserPasswordAuthProviderClient {
...
-    
-    func reset(password: String, withToken token: String, withTokenId tokenId: String, completionHandler: @escaping (Error?) -> Void) {
+
+    func reset(password: String,
+               withToken token: String,
+               withTokenId tokenId: String,
+               completionHandler: @escaping (Error?) -> Void) {
         dispatcher.run(withCompletionHandler: completionHandler) {
-            let _ = try super.reset(password: password, withToken: token, withTokenId: tokenId)
+            _ = try super.reset(password: password, withToken: token, withTokenId: tokenId)
```
jsflax: [nit][q] Same remark as above.

StitchCore-iOS/StitchCore-iOS/Core/Auth/Providers/UserPassword/UserPasswordAuthProvider.swift
```diff
@@ -88,45 +99,54 @@ public protocol UserPasswordAuthProviderClient {
...
         }
     }
-    
+
     func sendResetPasswordEmail(toEmail email: String, completionHandler: @escaping (Error?) -> Void) {
         dispatcher.run(withCompletionHandler: completionHandler) {
-            let _ = try super.sendResetPasswordEmail(toEmail: email)
+            _ = try super.sendResetPasswordEmail(toEmail: email)
```
jsflax: [nit][q] Same remark as above.

StitchCore-iOS/StitchCore-iOS/Core/Stitch.swift
jsflax: https://jira.mongodb.org/browse/STITCH-1356 I'm going to get on it once this PR is merged in.

StitchCore-iOS/StitchCore-iOS/Core/StitchAppClient.swift
```diff
@@ -39,7 +43,10 @@ public protocol StitchAppClient {
      *            passed in the `forProvider` parameter.
      */
     func serviceClient<T>(forService provider: AnyServiceClientProvider<T>) -> T
-    
+
+    // MARK: Functions
+
+    // swiftlint:disable line_length
```
jsflax: [minor] We should avoid disabling this check at all costs. The method sign below could have newline between parameters.

StitchCore/Common/Builder.swift
jsflax: This was a nested abstraction I had in core previously– Adam has simply moved it out to its own file.

With the builder pattern we are using, it saves us some repeated code for each builder, which we have a fair amount of. It's also the only way to create conformance with this kind of abstraction.

This was especially helpful with all of the builder extensions we wrote, based off of the Android builders.

The alternative is to do a more "old school" (less Swifty, more Java-y) builder pattern and do a fair amount of copy and pasting.

StitchCore/Sources/StitchCore/Auth/Providers/Anonymous/AnonymousCredential.swift
jsflax: Agreed. It would be nice if these were part of an enum.

StitchCore/Sources/StitchCore/Auth/Providers/UserPass/CoreUserPasswordAuthProviderClient.swift
jsflax: We should write a test for this. I believe it should be `authProviderRoute`, not `authProviderLoginRoute`, and that should clean this up.

StitchCore-iOS/StitchCore-iOS/Core/StitchAppClient.swift
edaniels: it's like that because jazzy cannot get the full function signature with newlines in protocols :(

jsflax: Ah wow, that seems like a pretty huge bug.

StitchCore-iOS/StitchCore-iOS/Core/Auth/Internal/StitchAuthImpl.swift
adamchel: I'll rename to `clientSupplier`. I think `Provider` is confusing because of the fact that it's for an authentication provider.

StitchCore-iOS/StitchCore-iOS/Services/Internal/StitchServiceImpl.swift
jsflax: I would consider a variadic list of ExtendedJSONRepresentables... we can also make the BSONArray class better.

StitchCore/Tests/stitch-ios-sdk-protoTests/Internal/Common/SynchronizedTests.swift
adamchel: It wasn't compiling without it but now when I remove it compiles so I'll just remove it.

StitchCore-iOS/StitchCore-iOS/Services/Internal/StitchServiceImpl.swift
adamchel: I just followed what we did in Java. Usage doesn't seem that bad:

```swift
stitchClient.callFunction(withName: "echoArg", withArgs: ["Hello world!"]) { (value, error) in
     // do stuff
}
```

But I'm okay with it if we can do the same in Java, and we can do it without messing with the completion handler pattern we've been using. (If the variadic parameter has to be the last parameter, then it would make the completion handler for callFunction different than every other asynchronous action.

StitchCore-iOS/StitchCore-iOS/Core/Auth/Providers/UserPassword/UserPasswordAuthProvider.swift
adamchel: these methods in super do have a return value (the raw `Response`), so we have to explicitly ignore it. (and it will properly throw when the response is not a 200-299)

StitchCore-iOS/StitchCore-iOS/Core/Auth/Providers/UserPassword/UserPasswordAuthProvider.swift
adamchel: same response as above

StitchCore-iOS/StitchCore-iOS/Core/Auth/Providers/UserPassword/UserPasswordAuthProvider.swift
adamchel: same response as above

StitchCore/Common/Builder.swift
adamchel: Going to leave as is then

StitchCore-iOS/StitchCore-iOS/Services/Internal/NamedServiceClientProvider.swift
adamchel: Not really sure. I didn't write this originally, and I figured we would make usability adjustments here once we actually implement some concrete services. This stuff isn't being used anywhere yet anyway.

PR #44, opened 2018-04-02
STITCH-1356 iOS SDK: Remove try notation for sync

jsflax:


StitchCore/Tests/stitch-ios-sdk-protoTests/Internal/Common/SynchronizedTests.swift
```diff
@@ -7,6 +6,7 @@ class SynchronizedTests: XCTestCase {
         fileprivate var value = 0
     }
 
+    @available(OSX 10.12, *)
```
adamchel: ah this was the same error I was getting that randomly disappeared when I upgraded to high sierra. I'm fine leaving this in here but is it fine that our test here won't run on linux?

StitchCore/Tests/stitch-ios-sdk-protoTests/Internal/Common/SynchronizedTests.swift
```diff
@@ -15,16 +15,16 @@ class SynchronizedTests: XCTestCase {
...
             for i in 0 ..< 100 {
                 queue.addOperation {
-                    try! sync(self) {
-                        let value = counter.value
-                        if i % 2 == 0 {
-                            usleep(50)
-                        }
-                        counter.value += 1
-                        XCTAssert(value == counter.value - 1)
+                    objc_sync_enter(self)
```
adamchel: [minor] Actually do we even need this test here anymore? Don't think this covers anything anymore except Foundation itself.

jsflax: Sort of a sanity check for me. Doesn't matter to me.

PR #45, opened 2018-04-05
STITCH-1354 Refactor error handling

adamchel:
I tried to refactor errors to be as consistent and cohesive as possible. It’s not too many changes, but I did think a lot about it as well as all of the edge cases. I’m happy to clarify anything

The gist of the change is that there now only three types of errors ever thrown by the SDK:
- `StitchClientError`: errors thrown before a request is made
- `StitchError`: errors thrown after a request is completed and a response is received
- Transport-specific errors: errors thrown while a request is being carried out (timeouts, could not reach server, etc.)

Main changes:
- StitchError now only represents errors that are from completed requests.
    - We now have `case unknownError` in `StitchError`
    - We no longer have `case requestError`. Instances were either replaced with an appropriate `StitchClientError`, or with an appropriate `unknownError` .
- StitchClientError now only represents errors that can occur before a request is ever made.
- Transports are responsible for throwing all errors that may be thrown while carrying out the request, and this is clarified in the comments.
- We no longer directly throw any error codes.
- `unknownError` and `invalidURL` were removed from `StitchErrorCode` as they aren’t actually Stitch error codes and are now their own appropriate cases in the other error enums.
- If the error returned from the server can’t be parsed as a JSON `ErrorCodable`, Stitch will return an unknown error with the response body text as the error message.

Drive-bys:
- XCTAssertThrowsError was not checking error messages as we thought, so I fixed all of our usages of it to actually verify that the correct error was being thrown
	- In the case of the roundTrip errors, they weren’t actually throwing Stitch errors at all, they were throwing errors specific to the transport, which is what we want
- Removed usages of `try! StitchCore.sync` and replaced with `objc_` stuff.

### DIFF
--- a/StitchCore/Sources/StitchCore/Internal/Net/FoundationHTTPTransport.swift
+++ b/StitchCore/Sources/StitchCore/Internal/Net/FoundationHTTPTransport.swift
@@ -56,7 +56,10 @@ public final class FoundationHTTPTransport: Transport {
 
         guard let response = finalResponse else {
             guard let err = error else {
-                throw StitchError.serviceError(withMessage: nil, withServiceErrorCode: .unknown)
+                throw StitchError.serviceError(
+                    withMessage: "no response from server",
+                    withServiceErrorCode: .unknown
+                )
             }
 
             throw err
--- a/StitchCore/Sources/StitchCore/Internal/Net/FoundationHTTPTransport.swift
+++ b/StitchCore/Sources/StitchCore/Internal/Net/FoundationHTTPTransport.swift
@@ -15,7 +15,7 @@ public final class FoundationHTTPTransport: Transport {
      */
     public func roundTrip(request: Request) throws -> Response {
         guard let url = URL(string: request.url) else {
-            throw StitchErrorCode.invalidURL
+            throw StitchClientError.missingURL
         }
 
         let defaultSession = URLSession(configuration: URLSessionConfiguration.default)
@@ -56,7 +56,7 @@ public final class FoundationHTTPTransport: Transport {
 
         guard let response = finalResponse else {
             guard let err = error else {
-                throw StitchErrorCode.unknown
+                throw StitchError.unknownError(withMessage: nil)
             }
 
             throw err
--- a/StitchCore/Sources/StitchCore/Internal/Net/FoundationHTTPTransport.swift
+++ b/StitchCore/Sources/StitchCore/Internal/Net/FoundationHTTPTransport.swift
@@ -15,7 +15,7 @@ public final class FoundationHTTPTransport: Transport {
      */
     public func roundTrip(request: Request) throws -> Response {
         guard let url = URL(string: request.url) else {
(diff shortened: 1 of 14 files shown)


adamchel:
Just modified the PR a little bit to address the concerns in the Slack discussion. There are now three types of errors, and all errors are under the singular `StitchError` enum. The comments in the StitchError enum describe the new structure.

StitchCore/Sources/StitchCore/Internal/Net/FoundationHTTPTransport.swift
```diff
@@ -56,7 +56,7 @@ public final class FoundationHTTPTransport: Transport {
 
         guard let response = finalResponse else {
             guard let err = error else {
-                throw StitchErrorCode.unknown
+                throw StitchError.serviceError(withMessage: nil, withServiceErrorCode: .unknown)
```
edaniels: [q] What's the case for this happening? Maybe we can have a better message?

jsflax: We should always be supplying some kind of message.

StitchCore/Sources/StitchCore/StitchError.swift
```diff
@@ -1,34 +1,51 @@
...
+     * error defined in the `StitchServiceErrorCode` enum.
+     *
+     * It is possible that the error code will be
+     * `StitchServiceErrorCode.unknown`, which can mean one of several possibilities: the Stitch server returned a
+     * message that this version of the SDK does not yet recognize, the server is not a Stitch server and returned an
+     * unexpected message, or the response was corrupted. In these cases, the associated message will be the plain text
+     * body of the response, or `nil` if the body is empty or not decodable as plain text.
      */
-    case serviceError(withMessage: String, withErrorCode: StitchErrorCode)
+    case serviceError(withMessage: String?, withServiceErrorCode: StitchServiceErrorCode)
```
jsflax: [minor] Message should never be optional

StitchCore/Sources/StitchCore/StitchError.swift
```diff
@@ -76,74 +93,85 @@ public enum StitchErrorCode: String, Codable, Error {
...
     /**
-     * Static utility method that accepts an HTTP `Response` object, and returns the `StitchError` contained within
-     * the response body. If there is no Stitch error in the body of the response, this will return
-     * `StitchErrorCode.unknown`.
+     * Static utility method that accepts an HTTP `Response` object, and returns the `StitchError` representing the
+     * the error in the response. If the error cannot be recognized, this will return a `StitchError.serviceError` with
+     * the `.unknown` error code.
      */
-    public static func handleRequestError(response: Response) -> Error {
+    public static func handleError(inResponse response: Response) -> StitchError {
```
jsflax: [nit] Should be `forResponse`

adamchel:
@edaniels @jsflax PTAL, along with the new Java PR.

PR #47, opened 2018-04-06
STITCH-1346 Add support for configuring data directory in StitchAppClient

adamchel:
Adds support for configuring an optional data directory for a Stitch app client. On iOS, this is set by default to the "Application Support" directory in the app's sandboxed filesystem. This was selected because of this description in the Apple docs:

"In general, this directory includes files that the app uses to run but that should remain hidden from the user. This directory can also include data files, configuration files, templates and modified versions of resources loaded from the app bundle."

The "Application Support" directory is also backed up by default.

### DIFF
--- a/StitchCore-iOS/StitchCore-iOS/Core/Stitch.swift
+++ b/StitchCore-iOS/StitchCore-iOS/Core/Stitch.swift
@@ -130,10 +130,12 @@ public class Stitch {
             finalConfigBuilder.storage = userDefaults
         }
 
-        // STITCH-1346:
-        //            if configBuilder.dataDirectory == nil {
-        //                finalConfigBuilder.dataDirectory = "/some/default/data/directory"
-        //            }
+        if configBuilder.dataDirectory == nil {
+            finalConfigBuilder.dataDirectory = try? FileManager.default.url(for: .applicationSupportDirectory,
+                                                                            in: .userDomainMask,
+                                                                            appropriateFor: nil,
+                                                                            create: true)
+        }
 
         if configBuilder.transport == nil {
             finalConfigBuilder.transport = FoundationHTTPTransport.init()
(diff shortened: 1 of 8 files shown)


StitchCore-iOS/StitchCore-iOS/Core/Stitch.swift
```diff
@@ -104,54 +105,57 @@ public class Stitch {
...
+                throw StitchInitializationError.userDefaultsFailure
             }
+            finalConfigBuilder.storage = userDefaults
+        }
 
-            if configBuilder.baseURL == nil {
-                finalConfigBuilder.baseURL = defaultBaseUrl
-            }
+        if configBuilder.dataDirectory == nil {
+            finalConfigBuilder.dataDirectory = try? FileManager.default.url(for: .applicationSupportDirectory,
```
edaniels: Can you add a test for this?

StitchCore/Sources/StitchCore/StitchClientConfiguration.swift
```diff
@@ -34,6 +41,11 @@ public struct StitchClientConfigurationImpl: StitchClientConfiguration, Buildee
      */
     public let baseURL: String
 
+    /**
+     * The local directory in which Stitch can store any data (e.g. embedded MongoDB data directory).
+     */
+    public let dataDirectory: URL?
```
jsflax: [q] Why is the optional? It seems like we can't really function without this.

adamchel: In the Android SDK it can be null, so kept it that way in iOS for consistency's sake. 

For now it's really only necessary when running embedded MongoDB, and for the server versions of the SDK we may make it fine not to configure it by default. I'm happy to make it non-optional though if we make the same change to Android. Thoughts @edaniels ?

edaniels: Should just make it non opt and validate

PR #53, opened 2018-05-01
STITCH-1257 Integrate Embedded MongoDB on iOS SDK

jsflax:


.gitignore
```diff
@@ -72,4 +75,4 @@ Carthage/Build
 fastlane/report.xml
 fastlane/Preview.html
 fastlane/screenshots
-fastlane/test_output
```
edaniels: [typo]?

StitchCore-iOS/StitchCore-iOS/Core/Internal/StitchAppClientImpl.swift
```diff
@@ -100,12 +100,12 @@ internal final class StitchAppClientImpl: StitchAppClient {
      * - returns: a service client whose type is determined by the `T` type parameter of the `AnyServiceClientProvider`
      *            passed in the `forProvider` parameter.
      */
-    public func serviceClient<T>(forService serviceClientProvider: AnyServiceClientProvider<T>) -> T {
-        return serviceClientProvider.client(
+    public func serviceClient<T>(forService serviceClientProvider: AnyServiceClientProvider<T>) throws -> T {
```
edaniels: [minor] This isn't going to be fun to use if this can throw. Can we avoid this somehow?

StitchMDBKit/Podfile
```diff
@@ -0,0 +1,10 @@
+# Uncomment the next line to define a global platform for your project
```
edaniels: [request] Could we name this module something else? I'd rather it break a convention and be `LocalMongoDBService`. Is that bad?

StitchMDBKit/StitchMDBKit/StitchMDBKit.h
```diff
@@ -0,0 +1,19 @@
+//
+//  StitchMDBKit.h
+//  StitchMDBKit
+//
+//  Created by Jason Flax on 4/30/18.
```
edaniels: [remove]

StitchMDBKit/StitchMDBKit/StitchMDBKit.h
```diff
@@ -0,0 +1,19 @@
+//
```
edaniels: [q] Is this file necessary because of pod?

StitchMDBKit/StitchMDBKit/LocalMongoDBService.swift
```diff
@@ -0,0 +1,168 @@
+import Foundation
```
edaniels: [q] Should these import be split into 1st, 2nd, 3rd party?

StitchMDBKitExample/StitchMDBKitExample/ToDoListViewController.swift
```diff
@@ -0,0 +1,132 @@
...
+
+        do {
+            // try to initialize the default stitch client
+            let client =
+                try Stitch.initializeDefaultAppClient(
+                    withConfigBuilder: StitchAppClientConfigurationBuilder.init {
+                        $0.clientAppId = "test-app"
+                    })
+            // try to get the default mongo client
+            let mongoClient = try client.serviceClient(forService: MongoDBService)
```
edaniels: [dope]

StitchMDBKitExample/StitchMDBKitExample/ToDoListViewController.swift
```diff
@@ -0,0 +1,132 @@
...
+
+/// Name of ToDoList db
+private let toDoListDatabaseName = "todo"
+/// Name of ToDoList collection
+private let toDoListCollectionName = "items"
+/// Cell reuse id (cells that scroll out of view can be reused)
+private let cellReuseIdentifier = "ToDoItemTableViewCell"
+
+/// View Controller for the ToDo List
+class ToDoListViewController: UITableViewController {
```
edaniels: I'm not reviewing this code :). Adam can if he wants.

.gitignore
jsflax: This is lower down in the file now

StitchMDBKit/Podfile
jsflax: I think that's fine. I played around with the `Kit` naming because it is more "iOS"-y. But I think it's more important that we have unification throughout the SDKs.

StitchMDBKit/StitchMDBKit/StitchMDBKit.h
jsflax: I believe it's so people can import us from obj-c projects.

StitchMDBKit/StitchMDBKit/LocalMongoDBService.swift
jsflax: I do not see any style guides that give any advice here, so it's up to us. We can do it that way.

StitchMDBKit/StitchMDBKit/StitchMDBKit.h
edaniels: So is it? This is new code.

StitchMDBKit/StitchMDBKit/LocalMongoDBService.swift
```diff
@@ -0,0 +1,168 @@
...
+private enum TrimMemoryCommand: String {
+    case mongoCommand = "trimMemory"
+    case aggressiveLevel = "aggressive"
+}
+
+/// Name of admin database to communicate commands to
+private let adminDatabaseName = "admin"
+
+/// Cached mongoClients associated to app ids
+private var localInstances = [String: MongoClient]()
```
adamchel: [nit] why is this a global rather than a static member of a singleton like it is in Java?

adamchel: The same could go for a lot of other stuff in this file.

StitchMDBKitExample/StitchMDBKitExample/ToDoListViewController.swift
```diff
@@ -0,0 +1,132 @@
...
+    private var toDoListCollection: MongoCollection!
+
+    override func viewDidLoad() {
+        super.viewDidLoad()
+
+        do {
+            // try to initialize the default stitch client
+            let client =
+                try Stitch.initializeDefaultAppClient(
```
adamchel: [nit] not that it matters all that much for this example, but if this eventually going to be a public example, this should probably be called in `application(didFinishLaunching...` with the mongomobile init code and here should be `getDefaultAppClient` since that will probably be the best practice. If this VC gets loaded multiple times (if for example someone modified the example so that this VC isn't the top-level VC), this would crash the app.

StitchMDBKit/StitchMDBKit/LocalMongoDBService.swift
jsflax: Because that's not a particularly Swifty way to do it. It's more expressive if, these static members are only used within this file, to have them at the file level.

StitchMDBKitExample/StitchMDBKitExample/ToDoListViewController.swift
jsflax: I was going by what we do in other examples. If we go with that approach, we should do the equivalent on Android as well.

jsflax:
@edaniels @adamchel ready for another look

LocalMongoDBService/LocalMongoDBService/LocalMongoDBService.swift
```diff
@@ -0,0 +1,168 @@
...
+        objc_sync_enter(self)
+        defer { objc_sync_exit(self) }
+
+        // if client is cached, return it
+        if let client = localInstances[clientInfo.clientAppId] {
+            return client
+        }
+
+        // else, create a new client
+        let settings = MongoClientSettings(dbPath: clientInfo.dataDirectory.path)
```
edaniels: This still needs to be configured to have the app id path concatted on like java: 
```
final String dbPath = String.format("%s/local_mongodb/0/", dataDir);
    final MongoClient client =
        MongoClients.create(MongoClientSettings.builder().dbPath(dbPath).build());
```

PR #57, opened 2018-05-31
STITCH-1540 Move Swift SDK to a mocking framework

adamchel:
This PR implements all of the unit tests from Java using a custom-built mocking framework that has some Mockito-like functionality, but it operates pretty differently from Mockito due to limitations in Swift. I've added the library as its own module with a single file, but when I have more time, I can flesh out the library into additional files and document it with the hopes of making it its own project.

To implement the unit tests, I had to make quite a few drive-by changes to fix issues and create overall parity between the Java and Swift SDKs. We are very close, and I will create tickets for the remaining tasks I didn't get to in this PR.


StitchCore/Sources/StitchCore/Internal/Net/FoundationHTTPTransport.swift
```diff
@@ -23,13 +23,9 @@ public final class FoundationHTTPTransport: Transport {
 
         let session = URLSession(configuration: sessionConfig)
 
-        var contentHeaders = request.headers
```
adamchel: I got rid of this since the doc request builders always add this when this is necessary, and we don't want to be adding this to requests with empty bodies.

MockUtils/Tests/LinuxMain.swift
```diff
@@ -0,0 +1,7 @@
+import XCTest
```
jsflax: [minor] Remove this file.

MockUtils/Tests/MockUtilsTests/MockUtilsTests.swift
```diff
@@ -0,0 +1,16 @@
+import XCTest
+@testable import MockUtils
+
+final class MockUtilsTests: XCTestCase {
+    func testExample() {
```
jsflax: [minor] remove this test

MockUtils/Tests/MockUtilsTests/XCTestManifests.swift
```diff
@@ -0,0 +1,9 @@
+import XCTest
+
+#if !os(macOS)
```
jsflax: [minor] remove this test

StitchCoreServicesTwilio/Tests/StitchCoreServicesTwilioTests/StitchCoreServicesTwilioTests.swift
```diff
@@ -1,6 +1,99 @@
 import XCTest
+import MockUtils
+import MongoSwift
+import StitchCore
 @testable import StitchCoreServicesTwilio
 
+final class MockCoreStitchService: CoreStitchService {
```
adamchel: I don't like how this is embedded directly in the test, or mixed into the test sources of StitchCoreTests. When I have more time (possibly after I'm done looking at Jason and Eric's PRs) I'll create a new target and product from the StitchCore package called StitchCoreMocks that will contain all of these in a way that can be imported without it being part of StitchCore or StitchCoreTests.

StitchCore/Sources/StitchCore/Auth/Internal/StitchAuthRequestClient.swift
```diff
@@ -10,13 +10,14 @@ public protocol StitchAuthRequestClient {
...
-     * authentication state, and should throw when not currently authenticated.
+     * Performs an authenticated request to the Stitch server, using the current authentication state, and should
+     * throw when not currently authenticated. Decodes the response body into a `Decodable` type based on the
+     * `DecodedT`  type parameter.
      *
-     * - returns: An `T` representing the response body as decoded JSON.
+     * - returns: The decoded body of the response.
      */
-    func doAuthenticatedJSONRequest<T: Decodable>(_ stitchReq: StitchAuthDocRequest) throws -> T
+    func doAuthenticatedRequest<DecodedT: Decodable>(_ stitchReq: StitchAuthRequest) throws -> DecodedT
```
jsflax: [nit] We shouldn't do named template types unless there are multiple types

StitchCore/Sources/StitchCore/Internal/Net/Requests/Request.swift
```diff
@@ -96,30 +137,32 @@ public struct Request: Buildee {
...
         self.method = method
         self.url = url
         self.timeout = timeout
+        self.headers = headers
+        self.body = body
+    }
 
-        self.headers = builder.headers ?? [:]
-        self.body = builder.body
+    public static func == (lhs: Request, rhs: Request) -> Bool {
```
jsflax: [style] extra space

StitchCore/Tests/StitchCoreTests/Auth/Internal/MockCoreStitchAuth.swift
```diff
@@ -0,0 +1,64 @@
...
+                        authRoutes: StubAuthRoutes.init(),
+                        storage: MemoryStorage(),
+                        startRefresherThread: false)
+        self.getAuthInfoMock.clearStubs()
+        self.getAuthInfoMock.clearInvocations()
+    }
+    
+    public var isLoggedInMock = FunctionMockUnit<Bool>()
+    override var isLoggedIn: Bool {
+        return isLoggedInMock.run()
```
jsflax: [q] These artifacts seem to defeat the purpose of mocking a bit. Is there any way we can avoid this much dependency injection?

StitchCore/Sources/StitchCore/Internal/Net/Requests/Request.swift
adamchel: This seems to be how it's done everywhere else, but I'm making the change if this is the style we want to follow.

StitchCore/Tests/StitchCoreTests/Auth/Internal/MockCoreStitchAuth.swift
adamchel: I don't think it completely defeats the purpose. We're wrapping a lot of functionality in the FunctionMockUnit, and IMO a little bit of test boilerplate is better than a lot of test boilerplate. This also seems to be the approach of other mocking libraries in Swift (including Cuckoo), it's just that those mocking libraries do code generation to generate the artifacts. In the long term we could look into code generation but I think this is fine for now.

PR #58, opened 2018-06-01
STITCH-1373 Implement AWS SES service in Core and iOS

adamchel:
This adds the AWS SES service with a bunch of drive-bys for the sake of Swift/Java parity, and to reorganize the `StitchCore` mock artifacts into their own target and product in `StitchCore` so that they can be used by more than just the `StitchCoreTests` target in `StitchCore`.


StitchCore-iOS/StitchCore-iOS/Services/Internal/ThrowingServiceClientFactory.swift
```diff
@@ -0,0 +1,42 @@
+import StitchCore
+
+/**
+ * A protocol describing a class that can provide clients for a particular named Stitch service.
+ */
+public protocol ThrowingServiceClientFactory {
```
jsflax: [q] where is this coming from? Hadn't I added this already in a previous commit?

StitchCoreServicesAwsSes-iOS/StitchCoreServicesAwsSes-iOSTests/StitchCoreServicesAwsSes_iOSTests.swift
```diff
@@ -0,0 +1,103 @@
...
+import StitchCoreAdminClient
+import StitchCoreTestUtils_iOS
+import StitchCoreServicesAwsSes
+@testable import StitchCoreServicesAwsSes_iOS
+
+class StitchCoreServicesAwsSes_iOSTests: BaseStitchIntTestCocoaTouch {
+    private let awsAccessKeyIdProp = "test.stitch.accessKeyId"
+    private let awsSecretAccessKeyProp = "test.stitch.secretAccessKey"
+    
+    private lazy var pList: [String: Any]? = {
```
jsflax: [minor] at this point, we should add a helper function for this in `BaseStitchIntTestCocoaTouch` and generecize it:

```swift
class BaseStitchIntTestCocoaTouch {
    func fetchPlist<T: BaseStitchIntTestCocoaTouch>(_ this: T.Type) -> [String: Any]? {
        let testBundle = Bundle(for: this)
        guard let url = testBundle.url(forResource: "Info", withExtension: "plist"),
            let myDict = NSDictionary(contentsOf: url) as? [String:Any] else {
                return nil
        }
        
        return myDict
    }
}

class StitchCoreServicesAwsSes_iOSTests {
    private lazy var pList: [String: Any]? = fetchPList(Self.self)
}
```

StitchCoreServicesAwsSes/Sources/StitchCoreServicesAwsSes/AwsSesSendResult.swift
```diff
@@ -0,0 +1,15 @@
+import Foundation
+
+/**
+ * The result of an AWS SES send request.
+ */
+public struct AwsSesSendResult: Decodable {
+    public enum CodingKeys: String, CodingKey {
```
jsflax: [minor] this is unnecessary if the keys are the same. We're better off having less code for these simple types

StitchCoreServicesAwsSes-iOS/StitchCoreServicesAwsSes-iOSTests/StitchCoreServicesAwsSes_iOSTests.swift
adamchel: 👍 Though the invocation that works seems to be `private lazy var pList: [String: Any]? = fetchPlist(type(of: self))` instead of using `Self.self.`

PR #59, opened 2018-06-04
STITCH-1372 Implement AWS S3 service in Core and iOS

adamchel:
Pretty much the same as SES, and also added some info to `contrib/README.md` that will help people add new modules until we have an automated script to do so.

This is slightly different from the Java implementation in that we expose additional overloads of all the putObject methods that allow specifying a timeout, since customizable timeouts on a per-request basis may be useful for people who are uploading large files via `putObject`. I will file a ticket for the same thing to be done on Java.

### DIFF
--- a/StitchCoreServicesAwsS3-iOS/StitchCoreServicesAwsS3-iOS/Internal/AwsS3ServiceClientImpl.swift
+++ b/StitchCoreServicesAwsS3-iOS/StitchCoreServicesAwsS3-iOS/Internal/AwsS3ServiceClientImpl.swift
@@ -1,0 +1,139 @@
+import Foundation
+import MongoSwift
+import StitchCore
+import StitchCore_iOS
+import StitchCoreServicesAwsS3
+
+public final class AwsS3ServiceClientImpl: AwsS3ServiceClient {
+    private let proxy: CoreAwsS3ServiceClient
+    private let dispatcher: OperationDispatcher
+    
+    internal init(withClient client: CoreAwsS3ServiceClient,
+                  withDispatcher dispatcher: OperationDispatcher) {
+        self.proxy = client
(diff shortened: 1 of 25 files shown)


StitchCoreServicesAwsS3/Sources/StitchCoreServicesAwsS3/Internal/CoreAwsS3ServiceClient.swift
```diff
@@ -0,0 +1,113 @@
...
+    
+    public init(withService service: CoreStitchServiceClient) {
+        self.service = service
+    }
+    
+    // An enumeration of the types that can be used in a put object request.
+    private enum PutObjectBody {
+        case string(String)
+        case bsonBinary(Binary)
+        case foundationData(Data)
```
jsflax: [minor] can just call this `data`, since Foundation is essentially the stdlib.

StitchCoreServicesAwsS3-iOS/StitchCoreServicesAwsS3-iOS/Internal/AwsS3ServiceClientImpl.swift
```diff
@@ -0,0 +1,139 @@
...
+            )
+        }
+    }
+    
+    public func putObject(bucket: String,
+                          key: String,
+                          acl: String,
+                          contentType: String,
+                          body: String,
+                          timeout: TimeInterval,
```
jsflax: [nit] should use default args for timeout, if possible

adamchel: as we discussed, this is not possible because protocols can't have default parameters

PR #60, opened 2018-06-05
STITCH-1371 Implement HTTP service in Core and iOS

adamchel:
almost no drivebys this time. `run_ios_tests` is still broken until we get them to reboot the box, but everything passes for me locally.

### DIFF
--- a/StitchCoreServicesHttp-iOS/StitchCoreServicesHttp-iOS/StitchCoreServicesHttp_iOS.h
+++ b/StitchCoreServicesHttp-iOS/StitchCoreServicesHttp-iOS/StitchCoreServicesHttp_iOS.h
@@ -1,0 +1,11 @@
+#import <UIKit/UIKit.h>
+
+//! Project version number for StitchCoreServicesHttp_iOS.
+FOUNDATION_EXPORT double StitchCoreServicesHttp_iOSVersionNumber;
+
+//! Project version string for StitchCoreServicesHttp_iOS.
+FOUNDATION_EXPORT const unsigned char StitchCoreServicesHttp_iOSVersionString[];
+
+// In this header, you should import all the public headers of your framework using statements like #import <StitchCoreServicesHttp_iOS/PublicHeader.h>
+
+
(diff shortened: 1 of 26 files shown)


StitchCoreServicesHttp-iOS/StitchCoreServicesHttp-iOS/StitchCoreServicesHttp_iOS.h
```diff
@@ -0,0 +1,11 @@
+#import <UIKit/UIKit.h>
+
+//! Project version number for StitchCoreServicesHttp_iOS.
+FOUNDATION_EXPORT double StitchCoreServicesHttp_iOSVersionNumber;
+
+//! Project version string for StitchCoreServicesHttp_iOS.
+FOUNDATION_EXPORT const unsigned char StitchCoreServicesHttp_iOSVersionString[];
+
+// In this header, you should import all the public headers of your framework using statements like #import <StitchCoreServicesHttp_iOS/PublicHeader.h>
```
jsflax: [nit] Can remove these newlines. We can probably start removing these headers as well.

StitchCoreServicesHttp/Sources/StitchCoreServicesHttp/HttpResponse.swift
```diff
@@ -0,0 +1,130 @@
...
+                }
+                
+                guard let value: String = try? document.get("value") else {
+                    throw MongoError.typeError(message: "expected string value for cookie value in HTTP service response")
+                }
+                
+                let keys = document.keys
+                
+                var path: String?
+                if keys.contains(HttpCookie.CodingKeys.path.rawValue) {
```
jsflax: [minor] Unless `Documents` operate radically differently than dictionaries (which they shouldn't given the try notation), you can remove all of these conditionals. Each one could be like:
`let path = try? document[HttpCookie.CodingKeys.path.rawValue]`.

adamchel: The syntax I was using is for the sake of type safety, but I played around with it more and made it cleaner like this:

```swift
let path = document[HttpCookie.CodingKeys.path.rawValue] as? String
```

StitchCoreServicesHttp-iOS/StitchCoreServicesHttp-iOS/StitchCoreServicesHttp_iOS.h
adamchel: Removed the newlines. Are these headers for the sake of Objective-C compatibility? Happy to remove them if we are definitely not supporting objc

PR #62, opened 2018-06-12
STITCH-1599 STITCH-1640 STITCH-1641 Reorganize modules, export symbols, and refactor callbacks

adamchel:
This PR encompasses three tickets, along with some extra drivebys:

**STITCH-1599 Be consistent with capitalization:**
The most significant change was changing (almost) every case of `Id` to `ID`. The places where I did not make the change were where we were mimicking the Swift driver (objectId, insertedId, upsertedId, insertedIds, etc.).

**STITCH-1640: Change callbacks to result enum pattern:**
Callbacks now accept a `ResultType<T>` enum that can either hold a result if an async operation succeeded, or a `StitchResult` otherwise.

**STITCH-1641: Export user-facing symbols:**
Every iOS-specific project now exports the user-facing symbols from its "core" "parent" (i.e. `StitchCore` imports `StitchCoreSDK`'s public symbols, and `StitchTwilioService` imports `StitchCoreTwilioService`'s symbols.

I also added an "uber" module as we've been calling it in the form of `StitchSDK`, which re-exports all of these symbols, as well as every user-facing symbol from the iOS-specific modules.

**Drive-bys:**
- Reorganized the xcworkspace into groups for easier code browsing in XCode
- Renamed every project in preparation for packaging
- Suffixed test files with their proper test type ("Unit" or "Int")
- updated some Runpath search paths in settings for the sake of getting tests building, but they're not consistent since it worked without runpaths for some tests and not others.

**Remaining work for packaging**
Just to help Jason when he works on packaging, I believe the following cleanup needs to be done before we can publish to Cocoapods

- Do something about the `.h` files in the iOS-specific modules. They all have old names and we should either rename them or get rid of them.
- Properly configure runpaths so tests can build and pass in a clean XCode environment
- Properly configure runpaths so the SDK can run on actual iOS devices.
- Change code and filenames so we don't have duplicate `UserAPIKeyAuthProviderClient` classes and files.


StitchAWSSESService/StitchAWSSESService/Exports.swift
```diff
@@ -0,0 +1,3 @@
+// Re-exported classes, structs, protocols, and enums from StitchCoreAWSS3Service
```
edaniels: s3 -> ses

PR #63, opened 2018-06-13
STITCH-1658 Swift SDK: Realign file structure to match XCode Groups

jsflax:


.gitignore
```diff
@@ -46,53 +46,55 @@ MockUtils/Package.resolved
...
-StitchCoreRemoteMongoDBService/Package.resolved
-StitchCoreRemoteMongoDBService/.build/
-StitchCoreRemoteMongoDBService/StitchCoreRemoteMongoDBService.xcodeproj/
-
-StitchCoreTwilioService/Packages/
-StitchCoreTwilioService/Package.pins
-StitchCoreTwilioService/Package.resolved
-StitchCoreTwilioService/.build/
-StitchCoreTwilioService/StitchCoreTwilioService.xcodeproj/
+Core/StitchCoreSDK/Packages/
```
edaniels: [q] Do we need to use the fully qualified name for these one they are in subdirs? i.e. can `StitchCoreSDK` become `sdk`?

jsflax: As discussed offline, this directory structure is the best we can do combined with our workspace.

PR #65, opened 2018-06-14
STITCH-1251 Implement FCM service and push service in Core and iOS

adamchel:
Super straightforward, ended up actually being simpler than Java because of `Codable`.


Core/Services/StitchCoreFCMService/Sources/StitchCoreFCMService/FCMSendMessageNotification.swift
```diff
@@ -0,0 +1,229 @@
...
+    public let color: String?
+    
+    /**
+     * The value of the badge on the home screen app icon. Note: for messages to iOS devices only.
+     */
+    public let badge: String?
+}
+
+
+public class FCMSendMessageNotificationBuilder {
```
jsflax: [nit][maybe] To be honest, all of these builders should probably be structs, so that we can take advantage of the `mutating` keyword. Of course, then we can't inherit from them, but it would be semantically more accurate. I'll leave this one up to you.

Core/Services/StitchCoreFCMService/Sources/StitchCoreFCMService/Internal/CoreFCMServiceClient.swift
```diff
@@ -0,0 +1,59 @@
...
+                            withRequest request: FCMSendMessageRequest) throws -> FCMSendMessageResult {
+        return try sendMessageInternal(
+            request: request,
+            targetTypeKey: SendFields.registrationTokensField.rawValue,
+            targetTypeValue: registrationTokens
+        )
+    }
+    
+    private func sendMessageInternal<T: BsonValue>(request: FCMSendMessageRequest,
+                                     targetTypeKey: String,
```
jsflax: [minor] This should be of type `SendFields` and we should take the `rawValue` within this method

Core/Services/StitchCoreFCMService/Sources/StitchCoreFCMService/Internal/CoreFCMServiceClient.swift
```diff
@@ -0,0 +1,59 @@
...
+        )
+    }
+    
+    private static let sendAction = "send"
+    
+    private enum SendFields: String {
+        // Target types
+        case userIDsField = "userIds"
+        case toField = "to"
+        case registrationTokensField = "registrationTokens"
```
jsflax: [nit] Can remove suffix `Field` as it's redundant

iOS/Services/StitchFCMService/StitchFCMService/FCMServicePushClient.swift
```diff
@@ -0,0 +1,45 @@
...
+     *
+     * - parameters:
+     *     - completionHandler: The completion handler to call when the user is deregistered or the operation fails.
+     *                          This handler is executed on a non-main global `DispatchQueue`.
+     */
+    func deregister(_ completionHandler: @escaping (StitchResult<Void>) -> Void)
+}
+
+public final class FCMPushService {
+    public static let sharedFactory =
```
jsflax: [minor] This is now the wrong format, and should be a top level constant

iOS/Services/StitchFCMService/StitchFCMService/StitchFCMService.h
```diff
@@ -0,0 +1,11 @@
+#import <UIKit/UIKit.h>
+
+//! Project version number for StitchFCMService.
+FOUNDATION_EXPORT double StitchFCMServiceVersionNumber;
+
+//! Project version string for StitchFCMService.
+FOUNDATION_EXPORT const unsigned char StitchFCMServiceVersionString[];
+
+// In this header, you should import all the public headers of your framework using statements like #import <StitchFCMService/PublicHeader.h>
```
jsflax: [nit] can remove some of these comments and extra lines

iOS/StitchCore/StitchCore/Core/Internal/StitchAppClientImpl.swift
```diff
@@ -16,6 +16,11 @@ internal final class StitchAppClientImpl: StitchAppClient {
         return _auth
     }
 
+    /**
+     * The client's underlying push notification component.
+     */
+    public var push: StitchPush
```
jsflax: [q][minor?] This doesn't seem correct. Why are we placing push in the StitchCore module?

iOS/StitchCore/StitchCore/Core/Push/StitchPush.swift
```diff
@@ -0,0 +1,16 @@
+import Foundation
+
+/**
+ * StitchPush can be used to get clients that can register for push notifications via Stitch.
+ */
+public protocol StitchPush {
```
jsflax: [q] Same as above, why is this in core?

Core/Services/StitchCoreFCMService/Sources/StitchCoreFCMService/FCMSendMessageNotification.swift
adamchel: agreed that it would be semantically nice, but as we discussed, this would break from the semantics of our other builders, so I'll leave as is.

Core/Services/StitchCoreFCMService/Sources/StitchCoreFCMService/Internal/CoreFCMServiceClient.swift
adamchel: :+1:

Core/Services/StitchCoreFCMService/Sources/StitchCoreFCMService/Internal/CoreFCMServiceClient.swift
adamchel: 👍

iOS/Services/StitchFCMService/StitchFCMService/FCMServicePushClient.swift
adamchel: 👍

iOS/Services/StitchFCMService/StitchFCMService/StitchFCMService.h
adamchel: :+1:

iOS/StitchCore/StitchCore/Core/Internal/StitchAppClientImpl.swift
adamchel: per offline discussion, leaving as is

iOS/StitchCore/StitchCore/Core/Push/StitchPush.swift
adamchel: per offline discussion, leaving as is

PR #103, opened 2018-11-14
STITCH-2211 Swift SDK: Create data models and implement local event coalescence

jsflax:
# "Port" the following classes:
## InstanceSynchronization:
- InstanceSynchronization now acts as a container for an internal config (`InstanceSynchronization.Config`). This was due to the fact that the current Java equivalent can exist in a partial state, which is a no-go in Swift. It also cleans up a significant amount of the encoding/decoding code, which was cumbersome. IMO, we should make these changes on Java as well.
- Conformance to `Sequence`. Can iterate through `NamespaceSynchronization.Configs`.
- Subscript access. `InstanceSynchronization` subscript uses a mutating getter to either get existing `NamespaceSynchronization`s or will create new ones.
## NamespaceSynchronization
- See above for container information.
- Conformance to `Sequence`. Can iterate through `CoreDocumentSynchronization.Configs`.
- Subscript access. `NamespaceSynchronization` subscript will either fetch existing `CoreDocumentSynchronization` by `id` or can set new ones by id.
## CoreDocumentSynchronization
- See above for container information. `Config` properties are now private. The properties are exposed in the outlying struct `CoreDocumentSynchronization` as computed properties. These getters/setters are synchronized via a `ReadWriteLock`, with the setters mutating the underlying `config`.
## ChangeEvent
- `OperationType` has been moved out to its own enum. 
    - The string values of each case are now implicit.
- `transform` (the method that converts a ChangeEvent with an associated type of `Document` to a user associated type) now takes on a more Swifty shape.
## UpdateDescription
- `UpdateDescription` has been moved out to its own class.
- Conversion to a `Document` is now computed lazily.
- `diff` is now a method extended to the `Document` class.
## ConflictHandler
- Uses associated types, which forces us to store locally using type erasure.
## ChangeEventListener
- Uses associated types, which forces us to store locally using type erasure.
## MongoNamespace
- Not sure why this class doesn't exist on `MongoMobile`. Simple port-over.
## DocumentVersionInfo
- `nextVersionDocument` is now lazily computed.
## ReadWriteLock
- Swift doesn't have a default RWLock, so this is a simple wrapper above `pthread_rwlock_t`.

## Other additions, notable or otherwise:
- More comments.
- More tests.
- Take advantage of nullability and mutability.
- More Swift.


Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/ChangeEvent.swift
```diff
@@ -0,0 +1,168 @@
...
+            documentKey: ["_id": docId],
+            updateDescription: nil,
+            hasUncommittedWrites: writePending)
+    }
+
+    /**
+     Generates a change event for a local update of a document in the given namespace referring
+     to the given document _id.
+
+     - parameter namespace: the namespace where the document was inserted.
```
adamchel: [nit] updated

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/ChangeEvent.swift
```diff
@@ -0,0 +1,168 @@
...
+            documentKey: ["_id": documentId],
+            updateDescription: update,
+            hasUncommittedWrites: writePending)
+    }
+
+    /**
+     Generates a change event for a local replacement of a document in the given namespace referring
+     to the given document _id.
+
+     - parameter namespace: the namespace where the document was inserted.
```
adamchel: [nit] replaced

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/ChangeEvent.swift
```diff
@@ -0,0 +1,168 @@
...
+            documentKey: ["_id": documentId],
+            updateDescription: nil,
+            hasUncommittedWrites: writePending)
+    }
+
+    /**
+     Generates a change event for a local deletion of a document in the given namespace referring
+     to the given document _id.
+
+     - parameter namespace: the namespace where the document was inserted.
```
adamchel: [nit] deleted

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/ChangeEvent.swift
```diff
@@ -0,0 +1,168 @@
...
+     - parameter documentId: the _id of the document that was updated.
```
adamchel: [nit] deleted

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/ChangeEvent.swift
```diff
@@ -0,0 +1,168 @@
...
+            updateDescription: update,
+            hasUncommittedWrites: writePending)
+    }
+
+    /**
+     Generates a change event for a local replacement of a document in the given namespace referring
+     to the given document _id.
+
+     - parameter namespace: the namespace where the document was inserted.
+     - parameter documentId: the _id of the document that was updated.
```
adamchel: [nit] replaced

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/NamespaceSynchronizationConfig.swift
```diff
@@ -0,0 +1,184 @@
...
+    }
+
+    /**
+     Configure a ConflictHandler and ChangeEventListener to this namespace.
+     These will be used to handle conflicts or listen to events, for this namespace,
+     respectively.
+     - parameter conflictHandler: a ConflictHandler to handle conflicts on this namespace
+     - parameter changeEventListener: a ChangeEventListener to listen to events on this namespace
+     */
+    mutating func configure<T: ConflictHandler, V: ChangeEventListener>(conflictHandler: T,
```
adamchel: [api discussion] In Java, this pattern works nicely, because you can pass an implementation of an interface inline (or use a lambda). In Swift, you can't really do that AFAIK. I know this isn't where we're going to expose configuration for the user, but have we thought about having configure variants that accept closures instead of protocols?

Core/Services/StitchCoreAWSS3Service/StitchCoreAWSS3Service.xcodeproj/project.pbxproj
```diff
@@ -10,8 +10,8 @@
 		492FA72D21909308001DFD53 /* StitchCoreSDK.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 492FA72C21909308001DFD53 /* StitchCoreSDK.framework */; };
 		492FA72F21909338001DFD53 /* MockUtils.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 492FA72E21909338001DFD53 /* MockUtils.framework */; };
 		492FA73321909351001DFD53 /* StitchCoreSDKMocks.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 492FA73221909351001DFD53 /* StitchCoreSDKMocks.framework */; };
-		AE3FAAFB31BE9D7D55FFC622 /* libPods-StitchCoreAWSS3Service.a in Frameworks */ = {isa = PBXBuildFile; fileRef = 471CBEEEDD51986508C0F591 /* libPods-StitchCoreAWSS3Service.a */; };
-		FDA5BEF356900B013C812B0F /* libPods-StitchCoreAWSS3ServiceTests.a in Frameworks */ = {isa = PBXBuildFile; fileRef = 85603B760FB854A36EB63122 /* libPods-StitchCoreAWSS3ServiceTests.a */; };
+		5CADF451D4021A34E0C51407 /* Pods_StitchCoreAWSS3Service.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = FCBD14893738BF6757F05B85 /* Pods_StitchCoreAWSS3Service.framework */; };
```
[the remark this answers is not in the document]
jsflax: Because we moved to frameworks with our podfile. Generally, any update to the podfile will cause pbxproj files to change.

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/BSONValue+Hashable.swift
```diff
@@ -0,0 +1,81 @@
...
+    case (let value as Double):
+        hasher.combine(value)
+    case (let value as Decimal128):
+        hasher.combine(value.data)
+    case (let value as Bool):
+        hasher.combine(value)
+    case (let value as String):
+        hasher.combine(value)
+    case (let value as RegularExpression):
+        hasher.combine(value.options)
```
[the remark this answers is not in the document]
jsflax: I wrote this based on how the driver wrote their equality function, but I can change it. Hasher just accumulates the hashes, it's nothing special.

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/ChangeEvent.swift
```diff
@@ -0,0 +1,168 @@
...
+     operation. This document may differ from the changes described in
+     updateDescription if other majority-committed operations modified
+     the document between the original update operation and the full
+     document lookup.
+     */
+    public let fullDocument: DocumentT?
+    /// The namespace (database and or collection) affected by the event.
+    public let ns: MongoNamespace
+    /**
+     The ObjectID of the document created or modified by the
```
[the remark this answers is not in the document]
jsflax: I lifted this right from the docs. https://docs.mongodb.com/manual/reference/change-events/

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/ChangeEvent.swift
```diff
@@ -0,0 +1,168 @@
...
+    public let documentKey: Document
+    /**
+     A document describing the fields that were updated or removed by
+     the update operation.
+     */
+    public let updateDescription: UpdateDescription?
+    /// Whether or not this ChangeEvent has pending writes.
+    public let hasUncommittedWrites: Bool
+
+    public static func == (lhs: ChangeEvent<DocumentT>, rhs: ChangeEvent<DocumentT>) -> Bool {
```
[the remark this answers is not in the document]
jsflax: I don't know why that would be the case. DocumentKey should be unique, and we shouldn't be comparing ChangeEvents between namespaces.

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/ChangeEvent.swift
```diff
@@ -0,0 +1,168 @@
...
+    /**
+     Transform a ChangeEvent of type Document to type T.
+
+     - parameter changeEvent: the ChangeEvent to transform
+     - returns: a ChangeEvent where the fullDocument is of type T
+     */
+    static func transform<T: Codable>(changeEvent: ChangeEvent<Document>) -> ChangeEvent<T> {
+        var fullDocument: T?
+        if let actualFullDocument = changeEvent.fullDocument {
+            fullDocument = try! BSONDecoder().decode(T.self,
```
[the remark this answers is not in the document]
jsflax: More of a [style] than a [nit].

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/NamespaceSynchronizationConfig.swift
jsflax: I think it's a good idea to do both. We can store it either way as their type erased equivalents, and the user can go about however they place. IMO, in an actual application, there's a stronger use case for having a singleton conform to these protocols than just a throwaway lambda. However, for the sake of the user, probably best to support both.

publish_pods.sh
```diff
@@ -0,0 +1,21 @@
+pod trunk push Core/StitchCoreSDK/StitchCoreSDK.podspec --allow-warnings --verbose
```
adamchel: [minor] if my docs stuff goes in before this, add `contrib/publish_docs.sh` before pushing any of the pods. that script will generate and publish the docs as long as you have the right AWS keys set up in your env

Core/Services/StitchCoreRemoteMongoDBService/Tests/StitchCoreRemoteMongoDBServiceTests/Sync/ChangeEventUnitTests.swift
```diff
@@ -0,0 +1,112 @@
...
+            fullDocument: document,
+            ns: namespace,
+            documentKey: ["_id": docId],
+            updateDescription: nil,
+            hasUncommittedWrites: writePending)
+
+        let actualLocalInsertEvent = ChangeEvent<Document>.changeEventForLocalInsert(namespace: namespace,
+                                                                                     document: document,
+                                                                                     writePending: writePending)
+        XCTAssertEqual(expectedLocalInsertEvent, actualLocalInsertEvent)
```
adamchel: since this uses the the incomplete `==` implementation for change events, this actually isn't comparing all the fields of the expected and actual change event

Core/Services/StitchCoreRemoteMongoDBService/Tests/StitchCoreRemoteMongoDBServiceTests/Sync/ChangeEventUnitTests.swift
```diff
@@ -0,0 +1,112 @@
...
+            operationType: .replace,
+            fullDocument: document,
+            ns: namespace,
+            documentKey: ["_id": docId],
+            updateDescription: nil,
+            hasUncommittedWrites: writePending)
+
+        let actualLocalReplaceEvent = ChangeEvent<Document>.changeEventForLocalReplace(namespace: namespace, documentId: docId, document: document, writePending: writePending)
+
+        XCTAssertEqual(expectedLocalReplaceEvent, actualLocalReplaceEvent)
```
adamchel: [nit] ditto

Core/Services/StitchCoreRemoteMongoDBService/Tests/StitchCoreRemoteMongoDBServiceTests/Sync/ChangeEventUnitTests.swift
```diff
@@ -0,0 +1,112 @@
...
+            hasUncommittedWrites: writePending)
+
+        let actualLocalUpdateEvent = ChangeEvent<Document>.changeEventForLocalUpdate(
+            namespace: namespace,
+            documentId: docId,
+            update: UpdateDescription.init(updatedFields: ["foo": 42, "bar": "baz"], removedFields: []),
+            fullDocumentAfterUpdate: document,
+            writePending: writePending)
+
+        XCTAssertEqual(expectedLocalUpdateEvent, actualLocalUpdateEvent)
```
adamchel: [nit] ditto

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/ChangeEvent.swift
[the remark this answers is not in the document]
dkaminsky: +1 adam

Prefer the more generic `_id` or, if you'd rather not use field name, "unique BSON identifier" or something along thos elines

Core/Services/StitchCoreRemoteMongoDBService/Tests/StitchCoreRemoteMongoDBServiceTests/Sync/ChangeEventUnitTests.swift
```diff
@@ -0,0 +1,112 @@
...
+        let expectedLocalDeleteEvent = ChangeEvent<Document>(
+            id: Document(),
+            operationType: .delete,
+            fullDocument: nil,
+            ns: namespace,
+            documentKey: ["_id": docId],
+            updateDescription: nil,
+            hasUncommittedWrites: writePending)
+
+        let actualLocalDeleteEvent = ChangeEvent<Document>.changeEventForLocalDelete(namespace: namespace,
```
adamchel: [nit] ditto

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/ChangeEvent.swift
dkaminsky: +1, should be [typo] and should be fixed

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/ChangeEvent.swift
dkaminsky: +1, should be [typo] and should be fixed

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/ChangeEvent.swift
dkaminsky: +1, should be [typo] and should be fixed

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/ChangeEvent.swift
dkaminsky: +1, should be [typo] and should be fixed

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/ChangeEvent.swift
dkaminsky: +1, should be [typo] and should be fixed

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/CoreDocumentSynchronizationConfig.swift
```diff
@@ -0,0 +1,338 @@
...
+                                       changeEvent: ChangeEvent<Document>) {
+        // if we were frozen
+        if (isPaused) {
+            // unfreeze the document due to the local write
+            isPaused = false
+            // and now the unfrozen document is now stale
+            isStale = true
+        }
+
+        docLock.writeLock()
```
dkaminsky: [major] shouldn't we acquire write lock before changing paused/stale status?

Core/Services/StitchCoreRemoteMongoDBService/Tests/StitchCoreRemoteMongoDBServiceTests/Sync/DocumentVersionInfoUnitTests.swift
```diff
@@ -0,0 +1,36 @@
+import XCTest
+import MongoMobile
+import MongoSwift
+@testable import StitchCoreRemoteMongoDBService
+
+class DocumentVersionInfoUnitTests: XCTestCase {
+    func testVersionDocument() {
```
adamchel: [minor] also add `testGetNextVersion` from the Android unit tests

Podfile
```diff
@@ -1,19 +1,22 @@
 workspace 'Stitch.xcworkspace'
```
dkaminsky: [q] Should this be committed and merged to master prior to the rest of this review, bypassing sync branch?

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/DocumentVersionInfo.swift
```diff
@@ -0,0 +1,159 @@
+import MongoSwift
```
dkaminsky: [nit] this wasn't in scope for this ticket IIRC

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/ChangeEvent.swift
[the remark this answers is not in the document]
adamchel: ah in that case the docs are wrong, I'll message Caleb

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/DocumentVersionInfo.swift
```diff
@@ -0,0 +1,159 @@
...
+    let filter: Document?
+    /**
+     Given a DocumentVersionInfo, returns a BSON document representing the next version. This means
+     and incremented version count for a non-empty version, or a fresh version document for
+     empty version.
+     */
+    lazy var nextVersion: Document = {
+        guard hasVersion,
+            let versionDoc = versionDoc,
+            let version = version else {
```
dkaminsky: [minor] isn't `hasVersion` redundant with `let version = version` based on the definition of hasVersion below?

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/BSONValue+Hashable.swift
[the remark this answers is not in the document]
adamchel: ah kk, no need to change then, esp. if this code is going away to do whatever the driver does

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/NamespaceSynchronizationConfig.swift
```diff
@@ -0,0 +1,184 @@
...
+            syncedDocuments.hash(into: &hasher)
+        }
+    }
+
+    typealias Element = CoreDocumentSynchronization
+    typealias Iterator = NamespaceSynchronizationIterator
+
+    /// Allows for the iteration of the document configs contained in this instance.
+    struct NamespaceSynchronizationIterator: IteratorProtocol {
+        typealias Element = CoreDocumentSynchronization
```
dkaminsky: [q] not redundant with wider-scoped typealias above?

Core/Services/StitchCoreRemoteMongoDBService/Tests/StitchCoreRemoteMongoDBServiceTests/CoreRemoteMongoClientUnitTests.swift
```diff
@@ -1,8 +1,10 @@
 import XCTest
 import Foundation
+import MongoMobile
 @testable import StitchCoreRemoteMongoDBService
 
-final class CoreRemoteMongoClientUnitTests: XCTestCase {
+final class CoreRemoteMongoClientUnitTests: XCMongoMobileTestCase {
```
dkaminsky: [q] this seems more a mongo mobile change not a sync hange?

Core/Services/StitchCoreRemoteMongoDBService/Tests/StitchCoreRemoteMongoDBServiceTests/Sync/ChangeEventUnitTests.swift
```diff
@@ -0,0 +1,112 @@
...
+                                                                                     documentId: docId,
+                                                                                     writePending: writePending)
+
+        XCTAssertEqual(expectedLocalDeleteEvent, actualLocalDeleteEvent)
+    }
+}
```
dkaminsky: [major] We should pin the error behavior for unknown with a test to make sure that all exception states are deliberate rather than implicit

Core/Services/StitchCoreRemoteMongoDBService/Tests/StitchCoreRemoteMongoDBServiceTests/Sync/XCMongoMobileTestCase.swift
```diff
@@ -0,0 +1,41 @@
+import Foundation
```
dkaminsky: [q] This seems more a mongo mobile thing than a sync thing?

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/CoreDocumentSynchronizationConfig.swift
jsflax: I ported over what happens on Android and assumed we had a reason for doing it this way. @edaniels?

jsflax:
Disregard that comment

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/ChangeEvent.swift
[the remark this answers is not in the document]
jsflax: Spoken about offline.

Core/Services/StitchCoreRemoteMongoDBService/StitchCoreRemoteMongoDBService.xcodeproj/StitchCoreRemoteMongoDBServiceTests_Info.plist
```diff
@@ -1,25 +1,26 @@
...
+	<key>CFBundlePackageType</key>
+	<string>BNDL</string>
+	<key>CFBundleShortVersionString</key>
+	<string>1.0</string>
+	<key>CFBundleSignature</key>
+	<string>????</string>
+	<key>CFBundleVersion</key>
+	<string>$(CURRENT_PROJECT_VERSION)</string>
+	<key>NSPrincipalClass</key>
+	<string>$(PRODUCT_NAME).XCMongoMobileConfiguration</string>
```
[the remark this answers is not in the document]
jsflax: This class is ran before and after test bundles are run. This is required to allow MongoMobile to start and teardown cleanly.

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/CoreDocumentSynchronizationConfig.swift
```diff
@@ -0,0 +1,338 @@
...
+internal func docFilter(forNamespace namespace: MongoNamespace,
+                        withDocumentId documentId: AnyBSONValue) -> Document {
+    return [
+        CoreDocumentSynchronization.Config.CodingKeys.namespace.rawValue: namespace.description,
+        CoreDocumentSynchronization.Config.CodingKeys.documentId.rawValue: documentId.value
+    ]
+}
+
+/**
+ The synchronization portal for this document.
```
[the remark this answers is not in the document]
jsflax: What would you say is a better choice of word? I had some trouble choosing.

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/CoreDocumentSynchronizationConfig.swift
```diff
@@ -0,0 +1,338 @@
...
+    }
+
+    /// The collection we are storing document configs in.
+    private let docsColl: MongoCollection<CoreDocumentSynchronization.Config>
+    /// Standard read-write lock.
+    private let docLock: ReadWriteLock = ReadWriteLock()
+    /// The configuration for this document.
+    private(set) var config: Config
+    /// The namespace this document is stored in.
+    var namespace: MongoNamespace { get { return config.namespace } }
```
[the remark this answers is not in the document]
jsflax: More of an explicit style choice

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/NamespaceSynchronizationConfig.swift
adamchel: when you implement a Swift protocol with associated types (the half-baked Swift feature that lets you do generic interfaces), it's generally idiomatic to use typealias to specify the type parameter of the protocol you're implementing

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/CoreDocumentSynchronizationConfig.swift
```diff
@@ -0,0 +1,338 @@
...
+                                  documentId: HashableBSONValue.init(documentId),
+                                  lastUncommittedChangeEvent: nil,
+                                  lastResolution: -1,
+                                  lastKnownRemoteVersion: nil,
+                                  isStale: false,
+                                  isPaused: false)
+    }
+
+    init(docsColl: MongoCollection<CoreDocumentSynchronization.Config>,
+         config: inout Config) {
```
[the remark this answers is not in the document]
jsflax: The configs should stay mutable so that the in memory mapping match what is in the local store.

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/MongoNamespace.swift
```diff
@@ -0,0 +1,32 @@
...
+    }
+
+    public init(databaseName: String, collectionName: String) {
+        self.databaseName = databaseName
+        self.collectionName = collectionName
+    }
+
+    public init(from decoder: Decoder) {
+        let description = try! decoder.singleValueContainer().decode(String.self)
+        let substrings = description.split(separator: ".")
```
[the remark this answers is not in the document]
jsflax: We'll have to make it a nullable initializer then.

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/NamespaceSynchronizationConfig.swift
```diff
@@ -0,0 +1,184 @@
...
+        /// a map of documents synchronized on this namespace, keyed on their documentIds
+        fileprivate var syncedDocuments: [HashableBSONValue: CoreDocumentSynchronization.Config]
+
+        init(namespace: MongoNamespace,
+             syncedDocuments: [HashableBSONValue: CoreDocumentSynchronization.Config]) {
+            self.namespace = namespace
+            self.syncedDocuments = syncedDocuments
+        }
+        static func == (lhs: NamespaceSynchronization.Config, rhs: NamespaceSynchronization.Config) -> Bool {
+            return true
```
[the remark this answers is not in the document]
jsflax: Oops

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/ReadWriteLock.swift
```diff
@@ -0,0 +1,38 @@
...
+        return pthread_rwlock_wrlock(&lock) == 0
+    }
+
+    @discardableResult
+    func tryWriteLock() -> Bool {
+        return pthread_rwlock_trywrlock(&lock) == 0
+    }
+
+    @discardableResult
+    func unlock() -> Bool {
```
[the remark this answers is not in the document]
jsflax: This was the question I brought up in Slack the other night when writing the RW lock. The unlock statement is smart enough to know the state of itself w.r.t read or write.

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/ChangeEvent.swift
```diff
@@ -0,0 +1,168 @@
+import MongoSwift
+
+/// The type of operation that occurred within a ChangeEvent.
+public enum OperationType: String, Codable {
+    case insert, delete, replace, update, unknown
```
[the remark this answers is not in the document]
jsflax: Invalidate?

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/ChangeEventListener.swift
```diff
@@ -0,0 +1,31 @@
+import MongoSwift
+
+/**
+ ChangeEventListener receives change event notifications.
+ */
+public protocol ChangeEventListener {
```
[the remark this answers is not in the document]
jsflax: That would be a choice we could make, but then we would remove the ability for users to configure using lambdas. They can do a switch statement over the operation type enum to replicate that same behavior.

Podfile
jsflax: We had to change dependencies for this PR.

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/DocumentVersionInfo.swift
jsflax: Good point, but that raises other questions of how this is setup.

Core/Services/StitchCoreRemoteMongoDBService/Tests/StitchCoreRemoteMongoDBServiceTests/CoreRemoteMongoClientUnitTests.swift
jsflax: We were doing this improperly before. Unfortunately, MongoMobile is setup in a way that it can only be initialized once and torndown once. So we need to be very careful about where we do those operations, especially with test bundles.

Core/Services/StitchCoreRemoteMongoDBService/Tests/StitchCoreRemoteMongoDBServiceTests/Sync/XCMongoMobileTestCase.swift
jsflax: See comment above.

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/CoreDocumentSynchronizationConfig.swift
jsflax: This is handled implicitly in the setters. I can make it explicit, however this is not a design flaw.

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/ChangeEvent.swift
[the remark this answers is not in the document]
adamchel: ah discussed with doug offline. `invalidate` is an event type that exists for MongoDB change events, but it doesn't exist for us since we close the client stream when we get an invalidate event on the Stitch server, so we don't need it. My other comment about the unknown event still stands though

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/CoreDocumentSynchronizationConfig.swift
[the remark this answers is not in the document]
adamchel: Maybe something along the lines of a "a class containing information about the synchronization state of a particular document."

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/NamespaceSynchronizationConfig.swift
jsflax: This will be exposed at a higher level.

Core/Services/StitchCoreRemoteMongoDBService/Tests/StitchCoreRemoteMongoDBServiceTests/Sync/DocumentVersionInfoUnitTests.swift
jsflax: I am technically testing the next version here.

jsflax:
@adamchel @dkaminsky I think we should kick the versioning changes downfield until we get to that ticket. I've laid down the base model which arguably was in scope for this one.

Otherwise, I've addressed your remarks. PTAL.

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/ChangeEvent.swift
```diff
@@ -2,13 +2,22 @@ import MongoSwift
 
 /// The type of operation that occurred within a ChangeEvent.
 public enum OperationType: String, Codable {
-    case insert, delete, replace, update, unknown
+    case insert = "INSERT"
```
adamchel: [minor] I think they should actually be lowercase

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/ChangeEvent.swift
```diff
@@ -2,13 +2,22 @@ import MongoSwift
...
 /// Represents possible fields that a change stream response document can have.
 public struct ChangeEvent<DocumentT: Codable>: Codable, Hashable {
+    enum CodingKeys: String, CodingKey {
+        case id = "_id", operationType, fullDocument, ns
+        case documentKey, updateDescription, hasUncommittedWrites
+    }
+
     /// Metadata related to the operation.
-    public let id: Document
+    public let id: AnyBSONValue
```
adamchel: [q] why did this change? this is going to always be a document

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/ChangeEvent.swift
```diff
@@ -44,12 +53,54 @@ public struct ChangeEvent<DocumentT: Codable>: Codable, Hashable {
...
+        self.hasUncommittedWrites = hasUncommittedWrites
+    }
+
+    public init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+
+        self.id = try container.decode(AnyBSONValue.self, forKey: .id)
+        self.operationType = OperationType.init(
+            rawValue: try container.decode(String.self, forKey: .operationType)
+        ) ?? .unknown
```
adamchel: [minor] should be `try?`

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/ConflictHandler.swift
```diff
@@ -19,25 +19,27 @@ public protocol ConflictHandler {
...
-                documentId: documentId,
-                localEvent: ChangeEvent<U.DocumentT>.transform(changeEvent: localEvent),
-                remoteEvent: ChangeEvent<U.DocumentT>.transform(changeEvent: remoteEvent))
-            return try! BSONEncoder().encode(documentT)
+            do {
+                let documentT = try conflictHandler.resolveConflict(
+                    documentId: documentId,
+                    localEvent: try ChangeEvent<U.DocumentT>.transform(changeEvent: localEvent),
+                    remoteEvent: try ChangeEvent<U.DocumentT>.transform(changeEvent: remoteEvent))
+                return try! BSONEncoder().encode(documentT)
```
adamchel: [minor] `try!` -> `try`

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/CoreDocumentSynchronizationConfig.swift
```diff
@@ -238,15 +264,15 @@ internal struct CoreDocumentSynchronization: Hashable {
 
      - parameter atVersion: the version for which the write as completed on
      */
-    mutating func setPendingWritesComplete(atVersion: Document) {
+    mutating func setPendingWritesComplete(atVersion: Document) throws {
         docLock.writeLock()
         defer { docLock.unlock() }
         self.lastUncommittedChangeEvent = nil
```
adamchel: [nit] not your code, but Doug and I spoke with Eric and we decided that we should drive-by rename `lastUncommittedChangeEvent` to `uncommittedChangeEvent` since the "last" is a misnomer. There is only ever one uncommitted change event since they're always coalesced.

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/InstanceSynchronizationConfig.swift
```diff
@@ -90,15 +98,23 @@ internal struct InstanceSynchronization: Sequence {
...
+            do {
+                let newConfig = try NamespaceSynchronization.init(namespacesColl: namespacesColl,
+                                                                  docsColl: docsColl,
+                                                                  namespace: namespace,
+                                                                  errorListener: errorListener)
+                try namespacesColl.insertOne(newConfig.config)
+                config.namespaces[namespace] = newConfig.config
+                return newConfig
+            } catch {
+                errorListener?.on(error: error, forDocumentId: nil)
```
adamchel: [discussion] hmm I see how this can be weird. An error on a nil document ID probably means some fatal error has occurred. 

This would definitely not be in scope for this ticket, but we should maybe think about exposing a fatal error handler to differentiate this from normal sync errors that might freeze a document.

Also to answer your offline question about how we should handle errors that happen as part of CRUD operations. I think sync Android operations just throw, and we wrap that as the exception of the Task. For Swift, we'll probably just wrap anything that throws as the error in the `StitchResult<T>`.

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/MongoNamespace.swift
```diff
@@ -2,31 +2,22 @@
  * A MongoDB namespace, which includes a database name and collection name.
  */
 public struct MongoNamespace: Codable, CustomStringConvertible, Hashable {
+    private enum CodingKeys: String, CodingKey {
```
adamchel: [minor] so I'm assuming this works, but then the underlying data structure for synchronization configs will be different on Android and iOS. I think it's important that we somehow maintain the "db.coll" String representation of the namespace.

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/NamespaceSynchronizationConfig.swift
```diff
@@ -26,8 +26,9 @@ internal struct NamespaceSynchronization: Sequence {
             self.namespace = namespace
             self.syncedDocuments = syncedDocuments
         }
+
         static func == (lhs: NamespaceSynchronization.Config, rhs: NamespaceSynchronization.Config) -> Bool {
-            return true
+            return lhs.namespace == rhs.namespace
```
adamchel: [q] when are we actually using the comparison? I think if we don't need it, we should get rid of it instead of having a shallow comparison

Core/Services/StitchCoreRemoteMongoDBService/Tests/StitchCoreRemoteMongoDBServiceTests/Sync/ChangeEventUnitTests.swift
```diff
@@ -107,6 +121,61 @@ class ChangeEventUnitTests: XCTestCase {
                                                                                      documentId: docId,
                                                                                      writePending: writePending)
 
-        XCTAssertEqual(expectedLocalDeleteEvent, actualLocalDeleteEvent)
+        try compare(expectedChangeEvent: expectedLocalDeleteEvent,
+                    to: actualLocalDeleteEvent)
+    }
+
+    func testRoundTrip() throws {
```
adamchel: [minor] can you add code to test the round trip for a change event with an unknown change event? It can be a `"blah"` event going in, and should be an `"unknown"` event coming out.

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/ChangeEvent.swift
jsflax: Are we sure?

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/ChangeEvent.swift
jsflax: If the operationType key should always be there, this should fail if the key doesn't exist.

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/ChangeEvent.swift
adamchel: https://github.com/mongodb/stitch-android-sdk/blob/master/core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/ChangeEvent.java#L110

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/MongoNamespace.swift
jsflax: Problem is, that's not how namespaces are represented by our API. I feel that API parity is more important.

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/ChangeEvent.swift
adamchel: Oh I see, this would only fail if `container.decode` fail if it's not a string. In that case that's fine

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/NamespaceSynchronizationConfig.swift
jsflax: We may use it down the line, later in our development process. I can remove it for now.

Core/Services/StitchCoreRemoteMongoDBService/Tests/StitchCoreRemoteMongoDBServiceTests/Sync/ChangeEventUnitTests.swift
jsflax: I already did.

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/MongoNamespace.swift
adamchel: where do we get a namespace back from our API?

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/NamespaceSynchronizationConfig.swift
adamchel: yeah let's remove it until we actually need it, and we can decide what makes two configs equal then

Core/Services/StitchCoreRemoteMongoDBService/Tests/StitchCoreRemoteMongoDBServiceTests/Sync/ChangeEventUnitTests.swift
adamchel: 🤦‍♂️  my bad

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/CoreDocumentSynchronizationConfig.swift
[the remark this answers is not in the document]
dkaminsky: +1 portal is a bad name -- is it a handle? a state manager? a state API?

Core/Services/StitchCoreRemoteMongoDBService/Tests/StitchCoreRemoteMongoDBServiceTests/Sync/ChangeEventUnitTests.swift
dkaminsky: I think this is resolved with the fixed == implementation

Core/Services/StitchCoreRemoteMongoDBService/Tests/StitchCoreRemoteMongoDBServiceTests/Sync/ChangeEventUnitTests.swift
dkaminsky: ditto

Core/Services/StitchCoreRemoteMongoDBService/Tests/StitchCoreRemoteMongoDBServiceTests/Sync/ChangeEventUnitTests.swift
dkaminsky: ditto

Core/Services/StitchCoreRemoteMongoDBService/Tests/StitchCoreRemoteMongoDBServiceTests/Sync/ChangeEventUnitTests.swift
dkaminsky: ditto

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/DocumentVersionInfo.swift
dkaminsky: which are..? :)

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/NamespaceSynchronizationConfig.swift
dkaminsky: kk, so even though it's the same typealias as above, this has some semantic meaning? or is it purely a style thing?

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/MongoNamespace.swift
jsflax: ChangeEvents, unless the docs are wrong

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/CoreDocumentSynchronizationConfig.swift
dkaminsky: +1 please fix now

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/MongoNamespace.swift
adamchel: ah it looks like in Android we decode them manuall in the change event decoder:

https://github.com/mongodb/stitch-android-sdk/blob/master/core/services/mongodb-remote/src/main/java/com/mongodb/stitch/core/services/mongodb/remote/sync/internal/ChangeEvent.java#L368

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/NamespaceSynchronizationConfig.swift
dkaminsky: not needed in tests?

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/DocumentVersionInfo.swift
jsflax: Questions that I stated we should kick downfield until the versioning ticket. Different semantics means there's a conversation worth being had about the original implementation.

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/MongoNamespace.swift
adamchel: I think keeping data the same on the device is going to be important as well, especially if we're going to end up going with a common C++ core. It helps us avoid doing local device data migrations in the future

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/ChangeEvent.swift
dkaminsky: Is it? You can't have a BSON String or Integer ID?

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/MongoNamespace.swift
adamchel: So I think we should still encode the MongoNamespace as `"\(db).\(coll)"` in the sync config

PR #107, opened 2018-11-20
STITCH-2008 - code complete ios changes to support location lookup fo…

dkaminsky:
…r client API calls - unit tests to follow

This follows the pattern established in the Android API by creating a specialized "app client" class that extends the existing functionality in order to retain the application's client app ID for the purpose of looking up the necessary application metadata from the Stitch server in order to know to which hostname to dispatch client API requests.

Tests will follow in a separate PR

### DIFF
--- a/Core/StitchCoreSDK/Sources/StitchCoreSDK/Internal/Net/AppMetadata.swift
+++ b/Core/StitchCoreSDK/Sources/StitchCoreSDK/Internal/Net/AppMetadata.swift
@@ -1,0 +1,24 @@
+import Foundation
+
+/// Application Metadata
+public final class AppMetadata: Codable, CustomStringConvertible {
+    let deploymentModel: String
+    let location: String
+    let hostname: String
+    
+    public var description: String {
+        let encoder = JSONEncoder()
+        
+        guard let data = try? encoder.encode(self),
+              let str = String(data: data, encoding: .utf8) else {
+            return "<json serialization error>"
+        }
+        return str
+    }
+    
+    enum CodingKeys: String, CodingKey {
+        case deploymentModel = "deployment_model"
+        case location
+        case hostname
+    }
+}
--- a/Core/StitchCoreSDK/Sources/StitchCoreSDK/Internal/Net/StitchAppRoutes.swift
+++ b/Core/StitchCoreSDK/Sources/StitchCoreSDK/Internal/Net/StitchAppRoutes.swift
@@ -4,10 +4,11 @@
 internal final class StitchAppRouteParts {
     static let baseRoute = "/api/client/v2.0"
     static let appRoute = baseRoute + "/app/%@"
+    static let appMetadataRoute = appRoute + "/location"
     static let functionCallRoute = appRoute + "/functions/call"
     static let baseAuthRoute = baseRoute + "/auth"
     static let baseAppAuthRoute = appRoute + "/auth"
-
+    
     static let sessionRoute = baseAuthRoute + "/session"
     static let profileRoute = baseAuthRoute + "/profile"
     static let authProviderRoute = baseAppAuthRoute + "/providers/%@"
@@ -23,27 +24,27 @@ public protocol StitchAuthRoutes {
      * The route on the server for getting a new access token.
      */
     var sessionRoute: String { get }
-
+    
     /**
      * The route on the server for fetching the currently authenticated user's profile.
      */
     var profileRoute: String { get }
-
+    
     /**
      * The base route on the server for authentication-related actions.
      */
     var baseAuthRoute: String { get }
-
+    
     /**
      * Returns the route on the server for getting information about a particular authentication provider.
      */
     func authProviderRoute(withProviderName providerName: String) -> String
-
+    
     /**
      * Returns the route on the server for logging in with a particular authentication provider.
      */
     func authProviderLoginRoute(withProviderName providerName: String) -> String
-
+    
     /**
      * Returns the route on the server for linking the currently authenticated user with an identity associated with a
      * particular authentication provider.
@@ -59,14 +60,20 @@ public final class StitchServiceRoutes {
      * The client app id of the app that these routes are for.
      */
     private let clientAppID: String
-
+    
     /**
      * Initializes these routes with the provided client app id.
      */
     fileprivate init(clientAppID: String) {
         self.clientAppID = clientAppID
     }
-
+    
+    /**
+     * Returns the route on the server for discovering application metadata.
+     */
+    public lazy var appMetadataRoute = String.init(format: StitchAppRouteParts.appMetadataRoute,
+                                              self.clientAppID)
+    
     /**
      * Returns the route on the server for executing a function.
      */
@@ -82,29 +89,29 @@ public struct StitchAppAuthRoutes: StitchAuthRoutes {
      * The client app id of the app that these routes are for.
      */
     private let clientAppID: String
-
+    
     /**
      * The route on the server for getting a new access token.
      */
     public var sessionRoute: String = StitchAppRouteParts.sessionRoute
-
+    
     /**
      * The route on the server for fetching the currently authenticated user's profile.
      */
     public var profileRoute: String = StitchAppRouteParts.profileRoute
-
+    
     /**
      * The route on the server for creating and modifying user API keys.
      */
     public var baseAuthRoute: String = StitchAppRouteParts.baseAuthRoute
-
+    
     /**
      * Initializes these routes with the provided client app id.
      */
     fileprivate init(clientAppID: String) {
         self.clientAppID = clientAppID
     }
-
+    
     /**
      * Returns the route on the server for getting information about a particular authentication provider.
      */
@@ -113,7 +120,7 @@ public struct StitchAppAuthRoutes: StitchAuthRoutes {
                            self.clientAppID,
                            providerName)
     }
-
+    
     /**
      * Returns the route on the server for logging in with a particular authentication provider.
      */
@@ -122,7 +129,7 @@ public struct StitchAppAuthRoutes: StitchAuthRoutes {
                            self.clientAppID,
                            providerName)
     }
-
+    
     /**
      * Returns the route on the server for linking the currently authenticated user with an identity associated with a
      * particular authentication provider.
@@ -142,7 +149,7 @@ public final class StitchAppRoutes {
      * Returns the authentication routes for the current app.
      */
     public let authRoutes: StitchAppAuthRoutes
-
+    
     /**
      * Returns the service routes for the current app.
      */
@@ -152,7 +159,7 @@ public final class StitchAppRoutes {
      * Returns the push routes for the current app.
      */
     public let pushRoutes: StitchPushRoutes
-
+    
     /**
      * Initializes the app routes with the provided client app id.
      */
@@ -162,3 +169,4 @@ public final class StitchAppRoutes {
         self.pushRoutes = StitchPushRoutes.init(clientAppID: clientAppID)
     }
 }
+
--- a/Core/StitchCoreSDK/Sources/StitchCoreSDK/Internal/Net/StitchRequestClient.swift
+++ b/Core/StitchCoreSDK/Sources/StitchCoreSDK/Internal/Net/StitchRequestClient.swift
@@ -7,9 +7,9 @@ import MongoSwift
 private func inspectResponse(response: Response) throws -> Response {
     guard response.statusCode >= 200,
         response.statusCode < 300 else {
-        throw StitchErrorCodable.handleError(forResponse: response)
+            throw StitchErrorCodable.handleError(forResponse: response)
     }
-
+    
     return response
 }
 
@@ -18,9 +18,23 @@ private func inspectResponse(response: Response) throws -> Response {
  */
 public protocol StitchRequestClient {
     /**
-     * Initializes the request client with the provided base URL and `Transport`.
+     * The base URL of the Stitch server to which this client will retrieve metadata.
      */
-    init(baseURL: String, transport: Transport, defaultRequestTimeout: TimeInterval)
+    var baseURL: String { get }
+    
+    /**
+     * The `Transport` which this client will use to make round trips to the Stitch server.
+     */
+    var transport: Transport { get }
+    
+    /**
+     * The number of seconds that a `Transport` should spend by default on an HTTP round trip before failing with an
+     * error.
+     *
+     * - important: If a request timeout was specified for a specific operation, for example in a function call, that
+     *              timeout should override this one.
+     */
+    var defaultRequestTimeout: TimeInterval { get }
 
     /**
      * Performs a request against the Stitch server with the given `StitchRequest` object.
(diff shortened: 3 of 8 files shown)


Core/StitchCoreSDK/Sources/StitchCoreSDK/Internal/Net/AppMetadata.swift
```diff
@@ -0,0 +1,24 @@
...
+    let deploymentModel: String
+    let location: String
+    let hostname: String
+    
+    public var description: String {
+        let encoder = JSONEncoder()
+        
+        guard let data = try? encoder.encode(self),
+              let str = String(data: data, encoding: .utf8) else {
+            return "<json serialization error>"
```
jsflax: [design] This should really be an impossible point to reach, but this block implies it is possible. For about the same amount of lines, you could just use a string or multiline string:

```
public lazy var description = """
        AppMetadata={
            deploymentModel=\(deploymentModel),
            location=\(location)
            hostname=\(hostname)
    }
"""
```

Since this class is virtually immutable, this value can also be lazily computed, since we don't need to recompute description after the first time.

As a bonus, here's a fancy but non-performant way to do descriptions as well:
```
public lazy var description: String = Mirror.init(
            reflecting: self
        ).children.reduce("AppMetadata={") { (res, child) -> String in
            "\(res)\(child.label!)=\(child.value), "
        }.dropLast(2) + "}"
```

Core/StitchCoreSDK/Sources/StitchCoreSDK/Internal/Net/AppMetadata.swift
```diff
@@ -0,0 +1,24 @@
+import Foundation
+
+/// Application Metadata
+public final class AppMetadata: Codable, CustomStringConvertible {
```
jsflax: [design] On second thought, this could be a `struct`: https://blog.usejournal.com/swift-basics-struct-vs-class-31b44ade28ae Regardless of whether it being a value vs. a reference affects our current intended use, I feel that `struct` as a keyword has an implication to the reader of what the structure may or may not be used for.

Core/StitchCoreSDK/Sources/StitchCoreSDK/Internal/Net/AppMetadata.swift
```diff
@@ -0,0 +1,24 @@
...
+        let encoder = JSONEncoder()
+        
+        guard let data = try? encoder.encode(self),
+              let str = String(data: data, encoding: .utf8) else {
+            return "<json serialization error>"
+        }
+        return str
+    }
+    
+    enum CodingKeys: String, CodingKey {
```
jsflax: [style] We generally place nested enums at the top of the structure they are nested in.

Core/StitchCoreSDK/Sources/StitchCoreSDK/Internal/Net/StitchAppRoutes.swift
```diff
@@ -4,10 +4,11 @@
 internal final class StitchAppRouteParts {
     static let baseRoute = "/api/client/v2.0"
     static let appRoute = baseRoute + "/app/%@"
+    static let appMetadataRoute = appRoute + "/location"
     static let functionCallRoute = appRoute + "/functions/call"
     static let baseAuthRoute = baseRoute + "/auth"
     static let baseAppAuthRoute = appRoute + "/auth"
-
+
```
jsflax: [style] extra tab

Core/StitchCoreSDK/Sources/StitchCoreSDK/Internal/Net/StitchAppRoutes.swift
```diff
@@ -23,27 +24,27 @@ public protocol StitchAuthRoutes {
      * The route on the server for getting a new access token.
      */
     var sessionRoute: String { get }
-
+
```
jsflax: [style] all the extra tabs

Core/StitchCoreSDK/Sources/StitchCoreSDK/Internal/Net/StitchRequestClient.swift
```diff
@@ -30,65 +44,156 @@ public protocol StitchRequestClient {
     func doRequest(_ stitchReq: StitchRequest) throws -> Response
 }
 
+extension StitchRequestClient {
+    func doRequest(_ stitchReq: StitchRequest, url: String) throws -> Response {
```
jsflax: [doc] Doc string please

Core/StitchCoreSDK/Sources/StitchCoreSDK/Internal/Net/StitchRequestClient.swift
```diff
@@ -30,65 +44,156 @@ public protocol StitchRequestClient {
...
+    }
+}
+
 /**
- * The implementation of `StitchRequestClient`.
+ * An implementation of `StitchRequestClient` that builds on `StitchRequestClientImpl` to
+ * add the ability to use a client application ID to target location-specific endpoints.
  */
-public final class StitchRequestClientImpl: StitchRequestClient {
+public class StitchAppRequestClientImpl: StitchRequestClient {
```
jsflax: [minor] Can finalize this class

Core/StitchCoreSDK/Sources/StitchCoreSDK/Internal/Net/StitchRequestClient.swift
```diff
@@ -30,65 +44,156 @@ public protocol StitchRequestClient {
...
+        let req = StitchRequest.init(path: self.appRoutes.serviceRoutes.appMetadataRoute,
+                                     method: Method.get, headers: [:],
+                                     timeout: self.defaultRequestTimeout, body: nil)
+        
+        let decoder = JSONDecoder()
         do {
-            response = try self.transport.roundTrip(request: self.buildRequest(stitchReq))
+            let response = try doRequest(req, url: self.baseURL)
+            guard let body = response.body else {
+                throw StitchError.requestError(withMessage: "empty body in location metadata", withRequestErrorCode: .decodingError)
```
jsflax: [style] Newline after the first comma

Core/StitchCoreSDK/Sources/StitchCoreSDK/Internal/Net/StitchRequestClient.swift
```diff
@@ -30,65 +44,156 @@ public protocol StitchRequestClient {
...
+            throw StitchError.requestError(withError: error, withRequestErrorCode: .bootstrapError)
         }
-        return try inspectResponse(response: response)
     }
+}
 
+/**
+ * The implementation of `StitchRequestClient`.
+ */
+public class StitchRequestClientImpl: StitchRequestClient {
```
jsflax: [minor] Can re-finalize this class

Core/StitchCoreSDK/Sources/StitchCoreSDK/Internal/Net/AppMetadata.swift
adamchel: [nit] similar to my comment on the Android PR, this class could be named `APIAppMetadata` if we want to more closely follow naming conventions for things that represent JSON coming from the API

Core/StitchCoreSDK/Sources/StitchCoreSDK/Internal/Net/StitchAppRoutes.swift
adamchel: [meta q] mostly a question for Jason, but is our build process in a stable enough state that we can do the add linter build phases ticket (STITCH-1597)? It was previously in limbo because SPM would overwrite build phases when generating the xcodeproj but now that all of our Xcode projects are checked into the repo and cocoapods doesn't overwrite them, adding the linter phases should be trivial (aside from actually getting our code to pass the linter)

Core/StitchCoreSDK/Sources/StitchCoreSDK/Internal/Net/StitchRequestClient.swift
```diff
@@ -7,9 +7,9 @@ import MongoSwift
 private func inspectResponse(response: Response) throws -> Response {
     guard response.statusCode >= 200,
         response.statusCode < 300 else {
-        throw StitchErrorCodable.handleError(forResponse: response)
+            throw StitchErrorCodable.handleError(forResponse: response)
```
adamchel: [nit] I think this was previously at the correct indentation level. the guard statement could probably fit on one line making that clearer.

Core/StitchCoreSDK/Sources/StitchCoreSDK/StitchError.swift
```diff
@@ -24,8 +24,18 @@ public enum StitchError: Error {
      * thrown when attempting to decode the response. An error code is included, which indicates whether the error
      * was a transport error or decoding error.
      */
-    case requestError(withError: Error, withRequestErrorCode: StitchRequestErrorCode)
+    case requestError(withMessage: String, withRequestErrorCode: StitchRequestErrorCode)
```
adamchel: [neat] I did not know you could have multiple enum cases with the same name, but different types of associated values.

Core/StitchCoreSDK/Sources/StitchCoreSDK/Internal/Net/StitchRequestClient.swift
```diff
@@ -30,65 +44,156 @@ public protocol StitchRequestClient {
...
 /**
- * The implementation of `StitchRequestClient`.
+ * An implementation of `StitchRequestClient` that builds on `StitchRequestClientImpl` to
+ * add the ability to use a client application ID to target location-specific endpoints.
  */
-public final class StitchRequestClientImpl: StitchRequestClient {
+public class StitchAppRequestClientImpl: StitchRequestClient {
     /**
-     * The base URL of the Stitch server to which this client will make requests.
+     * The base URL of the Stitch server to which this client will retrieve metadata.
```
adamchel: [grammar] "to which" -> "from which"

Core/StitchCoreSDK/Sources/StitchCoreSDK/Internal/Net/AppMetadata.swift
adamchel: I see your point, but idk if I agree. The first non-reflection approach has to change if we ever add more fields to app metadata, and in the case we do add fields that can't be serialized, we would want the description to fail with a JSON serialization error rather than omit then entirely because we forgot to update the `description` getter

adamchel: so while the case is technically impossible to reach for now, it could become possible with future changes that might not touch that particular method.

jsflax: There's a few reasons why my first example is the most efficient and most idiomatic approach. Ranked by importance:
- `CustomStringConvertible` is the canonical representation of some instance. Think about this protocol more in the context of an `asString` method as opposed to a `toString` method. An `ObjectId` would be canonised as its hex string value, not the extended JSON value. Because as a String, that's what it actually is. A `Point` class would be `(\(x), \(y))`.
- This struct is intended to be immutable. It should be lazily computed to avoid unnecessary computation.
- The reason we are encountering all of these weird `try` issues around getters/setters is because we're abusing the concept of computed properties. They are still supposed to be simple properties– doing computations that can `throw` breaches the intended use of computed properties, especially if you need to resurface the error. Acceptable use cases would be a default value if a `try` fails.
- Serialising to JSON every time you want to get the string description is unnecessarily cumbersome.

If we add fields that cannot be serialised (which should never be possible, we are the one that chooses what can or cannot be serialised), failing in the `description` is absolutely the wrong place to be failing. If we add fields that can be serialised and are added to the description, forgetting to update the `description` getter is an unacceptable reason to break all of the rules I've listed above. The same level of developer responsibility would apply to Java's `toString`, which generally does not use JSON or reflection.

Core/StitchCoreSDK/Sources/StitchCoreSDK/Internal/Net/StitchAppRoutes.swift
jsflax: Yes.

Core/StitchCoreSDK/Sources/StitchCoreSDK/Internal/Net/AppMetadata.swift
adamchel: fair enough, all this makes sense

dkaminsky: removed for now

Core/StitchCoreSDK/Sources/StitchCoreSDK/Internal/Net/AppMetadata.swift
```diff
@@ -0,0 +1,24 @@
+import Foundation
+
+/// Application Metadata
+public final class AppMetadata: Codable, CustomStringConvertible {
+    let deploymentModel: String
```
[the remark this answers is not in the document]
dkaminsky: added

Core/StitchCoreSDK/Sources/StitchCoreSDK/Internal/Net/AppMetadata.swift
dkaminsky: done

Core/StitchCoreSDK/Sources/StitchCoreSDK/Internal/Net/AppMetadata.swift
dkaminsky: done

Core/StitchCoreSDK/Sources/StitchCoreSDK/Internal/Net/StitchAppRoutes.swift
dkaminsky: fixed

Core/StitchCoreSDK/Sources/StitchCoreSDK/Internal/Net/StitchAppRoutes.swift
dkaminsky: fixed

Core/StitchCoreSDK/Sources/StitchCoreSDK/Internal/Net/StitchRequestClient.swift
dkaminsky: done

Core/StitchCoreSDK/Sources/StitchCoreSDK/Internal/Net/StitchRequestClient.swift
dkaminsky: done

Core/StitchCoreSDK/Sources/StitchCoreSDK/Internal/Net/StitchRequestClient.swift
dkaminsky: done

Core/StitchCoreSDK/Sources/StitchCoreSDK/Internal/Net/StitchRequestClient.swift
dkaminsky: done

Core/StitchCoreSDK/Sources/StitchCoreSDK/StitchError.swift
dkaminsky: 👍

Core/StitchCoreSDK/Sources/StitchCoreSDK/Internal/Net/StitchRequestClient.swift
dkaminsky: fixed

Core/StitchCoreSDK/Sources/StitchCoreSDK/Internal/Net/AppMetadata.swift
dkaminsky: fixed

Core/StitchCoreSDK/Sources/StitchCoreSDK/Internal/Net/StitchRequestClient.swift
dkaminsky: fixed

dkaminsky:
> You will add the tests before closing the ticket though, right?

Yep

PR #114, opened 2018-12-03
STITCH-2215 Swift SDK: Implement configuration for the Sync interface

jsflax:
Most of `configure` we got for free over the past several PRs for sync.

This adds `configure` for blocks and all of the tests that required configure to be hooked up all the way.


Darwin/Services/StitchRemoteMongoDBService/StitchRemoteMongoDBService/Sync.swift
```diff
@@ -15,16 +15,33 @@ public class Sync<DocumentT: Codable> {
...
      - parameter conflictHandler: the conflict resolver to invoke when a conflict happens between local
      and remote events.
-     - parameter changeEventListener: the event listener to invoke when a change event happens for the
+     - parameter changeEventDelegate: the event listener to invoke when a change event happens for the
      document.
      - parameter errorListener: the error listener to invoke when an irrecoverable error occurs
      */
-    func configure<CH: ConflictHandler, CEL: ChangeEventListener>(
+    func configure(
+        conflictHandler: @escaping BlockConflictHandler<DocumentT>.ResolveConflictBlock,
```
adamchel: [style minor] I think for the sake of readability and API documentation, we should avoid using a typealias and just use the full block type. We don't use type aliases anywhere else for block arguments  It will be easier to understand, and it will let us avoid making `BlockConflictHandler` a `public` type. It can be `internal` instead.

Darwin/Services/StitchRemoteMongoDBService/StitchRemoteMongoDBServiceTests/CallbackJoiner.swift
```diff
@@ -0,0 +1,98 @@
+import Foundation
+import StitchCore
+import XCTest
+import MongoSwift
+@testable import StitchCoreRemoteMongoDBService
+
+private class SynchronizedDispatchDeque {
```
adamchel: [nit] The classes in this file feel like they should belong in `StitchDarwinCoreTestUtils`. It seems like it would be very useful for all of our async tests. We really abused `XCTestExpectation` in other places and this could really simplify future tests. Is the `@testable` import of the remote mongo service actually necessary for anything here?

adamchel: ah I see that it's for the `ReadWriteLock` Honestly the `ReadWriteLock` should probably go into `StitchCore` since it doesn't actually use any remote mongo service constructs, and then this class can be decoupled from remote mongo tests

Darwin/Services/StitchRemoteMongoDBService/StitchRemoteMongoDBServiceTests/RemoteMongoClientIntTests.swift
```diff
@@ -15,8 +16,8 @@ class RemoteMongoClientIntTests: BaseStitchIntTestCocoaTouch {
     
     private lazy var mongodbUri: String = pList?[mongodbUriProp] as? String ?? "mongodb://localhost:26000"
 
-    private let dbName = ObjectId().description
-    private let collName = ObjectId().description
+    private let dbName = "dbName"
```
adamchel: [q] why did this change?

Darwin/Services/StitchRemoteMongoDBService/StitchRemoteMongoDBServiceTests/RemoteMongoClientIntTests.swift
```diff
@@ -26,6 +27,17 @@ class RemoteMongoClientIntTests: BaseStitchIntTestCocoaTouch {
         try! prepareService()
     }
 
+    override func tearDown() {
+        let joiner = CallbackJoiner()
+        getTestColl().deleteMany([:], joiner.capture())
```
adamchel: [nice] This is soooo much nicer than XCTestExpectation, I wish we had this before

Darwin/Services/StitchRemoteMongoDBService/StitchRemoteMongoDBServiceTests/RemoteMongoClientIntTests.swift
```diff
@@ -1019,16 +1031,281 @@ class RemoteMongoClientIntTests: BaseStitchIntTestCocoaTouch {
...
+        sync.count(joiner.capture())
+        XCTAssertEqual(0, joiner.value())
+
+        let doc1 = ["hello": "world", "a": "b"] as Document
+        let doc2 = ["hello": "computer", "a": "b"] as Document
+        sync.insertMany(documents: [doc1, doc2], joiner.capture())
+        sync.count(joiner.capture())
+
+        XCTAssertEqual(2, joiner.value())
+        // TODO: STITCH-2262 Add delete case
```
adamchel: [nit] pull this TODO out into `testSync_Delete*()` function stubs, this is an odd place for the TODO

Darwin/Services/StitchRemoteMongoDBService/StitchRemoteMongoDBServiceTests/CallbackJoiner.swift
```diff
@@ -0,0 +1,98 @@
...
+
+    func removeFirst() -> DispatchWorkItem {
+        rwLock.writeLock()
+        defer { rwLock.unlock() }
+        return workItems.removeFirst()
+    }
+}
+
+/// Synchronously capture the value of a asynchronous callback.
+class CallbackJoiner {
```
adamchel: [nit] `CallbackJoiner` could maybe be called `CallbackAwaiter`, since we use `await` in the other SDKs, but I'm also fine keeping it as is

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/NamespaceSynchronizationConfig.swift
```diff
@@ -180,21 +180,21 @@ internal struct NamespaceSynchronization: Sequence {
     }
 
     /**
-     Configure a ConflictHandler and ChangeEventListener to this namespace.
+     Configure a ConflictHandler and ChangeEventDelegate to this namespace.
      These will be used to handle conflicts or listen to events, for this namespace,
      respectively.
 
      TODO STITCH-2212: Add typealias lambdas to the higher level call for this function.
```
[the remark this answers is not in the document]
jsflax: That's a nit, not a minor. This is just a comment I hadn't removed.

Darwin/Services/StitchRemoteMongoDBService/StitchRemoteMongoDBServiceTests/RemoteMongoClientIntTests.swift
jsflax: The database name for sync ended up being too long and mongo was rejecting it.

Darwin/Services/StitchRemoteMongoDBService/StitchRemoteMongoDBServiceTests/RemoteMongoClientIntTests.swift
jsflax: As an integration test, I want to delete the documents and test the count has moved down to zero.

Darwin/Services/StitchRemoteMongoDBService/StitchRemoteMongoDBServiceTests/CallbackJoiner.swift
jsflax: I agree.

jsflax: I'm going to swap this in the next PR, where I'll need the RWLock actually moved.

Darwin/Services/StitchRemoteMongoDBService/StitchRemoteMongoDBServiceTests/CallbackJoiner.swift
adamchel: fair enough, I'm happy with `CallbackJoiner`

Darwin/Services/StitchRemoteMongoDBService/StitchRemoteMongoDBServiceTests/RemoteMongoClientIntTests.swift
adamchel: really? that seems odd. it seems like "db.collName" is allowed to be 120 bytes. Is something else making the name longer? is `.description` actually the hex value of the db? this could become a problem when we are namespacing by user id and client app id

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/CoreDocumentSynchronizationConfig.swift
```diff
@@ -37,15 +37,15 @@ internal struct CoreDocumentSynchronization: Hashable {
         let namespace: MongoNamespace
         let documentId: HashableBSONValue
         fileprivate var uncommittedChangeEvent: ChangeEvent<Document>?
-        fileprivate var lastResolution: UInt
+        fileprivate var lastResolution: UInt32
```
adamchel: [minor] this should be `UInt64`. It's a `long` in Java.

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/DataSynchronizer.swift
```diff
@@ -74,7 +74,7 @@ public class DataSynchronizer: NetworkStateDelegate, FatalErrorListener {
     /// The user's error listener
     private var errorListener: ErrorListener?
     /// Current sync pass iteration
-    private var logicalT: UInt = 0
+    private var logicalT: UInt32 = 0
```
adamchel: [minor] `UInt64`

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/DataSynchronizer.swift
```diff
@@ -870,11 +870,11 @@ public class DataSynchronizer: NetworkStateDelegate, FatalErrorListener {
     }
 
     internal static func localConfigDBName(withInstanceKey instanceKey: String) -> String {
-        return "sync-config-\(instanceKey)"
+        return "sync_config_\(instanceKey)"
```
adamchel: :+1:

Core/Services/StitchCoreRemoteMongoDBService/Tests/StitchCoreRemoteMongoDBServiceTests/Sync/CoreDocumentSynchronizationConfigTests.swift
```diff
@@ -27,7 +27,7 @@ class CoreDocumentSynchronizationConfigTests: XCMongoMobileTestCase {
         let isPaused = true
         let isStale = true
         let lastKnownRemoteVersion = DocumentVersionInfo.freshVersionDocument()
-        let lastResolution = 42.0
+        let lastResolution: UInt32 = 42
```
adamchel: [minor] `UInt64`

Darwin/Services/StitchRemoteMongoDBService/StitchRemoteMongoDBServiceTests/RemoteMongoClientIntTests.swift
jsflax: The description is just the odd. I think client app id is short enough that it shouldn't be a problem, but yes, user id may cause issues. We may want to rename the sync databases to their most possible abbreviated forms (sync_config => sc) before release.

adamchel: What I was saying is it's unusual that the oid by itself made the collection name too long. I wonder if there's something weird that's making the limit smaller than it should be. What's the exact error you get when leave the db and coll name as the object ID? sync_config_5b75c94c81941e2ad865a475.5b75c94c81941e2ad865a475 is only 62 bytes and the limit should be 120

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/DataSynchronizer.swift
```diff
@@ -875,6 +875,6 @@ public class DataSynchronizer: NetworkStateDelegate, FatalErrorListener {
     
     internal static func localUserDBName(withInstanceKey instanceKey: String,
                                          for namespace: MongoNamespace) -> String {
-        return "sync_user_\(instanceKey)_\(namespace.databaseName)"
+        return "sync_user_\(instanceKey)-\(namespace.databaseName)"
```
dkaminsky: why?

PR #115, opened 2018-12-06
STITCH-2262 Happy path DELETE

adamchel:
These are the implementation and tests for happy path delete. Also drove-by a fix to make the logicalT and lastResolution fields be `Int64` so they are properly encoded. This is not quite ready to be merged, since I believe I will need to add a few more tests once Jason merges STITCH-2215

### DIFF
--- a/Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/DataSynchronizer.swift
+++ b/Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/DataSynchronizer.swift
@@ -3,7 +3,7 @@ import MongoSwift
 import MongoMobile
 import StitchCoreSDK
 
-/// Internal extension so we can initialize using
+/// Internal extensions so we can initialize using
 /// internal initializer.
 extension UpdateResult {
     init(matchedCount: Int,
@@ -14,6 +14,12 @@ extension UpdateResult {
         self.modifiedCount = modifiedCount
         self.upsertedId = upsertedId
         self.upsertedCount = upsertedCount
+    }
+}
+
+extension DeleteResult {
+    init(deletedCount: Int) {
+        self.deletedCount = deletedCount
     }
 }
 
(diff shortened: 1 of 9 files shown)


Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/DataSynchronizer.swift
```diff
@@ -17,6 +17,12 @@ extension UpdateResult {
     }
 }
 
+extension DeleteResult {
```
jsflax: [swift5] We will need to remove these to support Swift 5. We should probably use the same strategy we used on Android of creating our own models (SyncDeleteResults, etc.). I'll file a ticket for it.

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/DataSynchronizer.swift
```diff
@@ -507,8 +551,57 @@ public class DataSynchronizer: NetworkStateDelegate, FatalErrorListener {
...
+                withClientErrorCode: StitchClientErrorCode.couldNotLoadSyncInfo)
+        }
+        let lock = nsConfig.nsLock
+        lock.writeLock()
+        defer { lock.unlock() }
+
+        var eventsToEmit = [(BSONValue, ChangeEvent<Document>)]()
+        let localColl = try localCollection(for: namespace, withType: Document.self)
+
+        let idsToDelete = try localColl.find(filter).map { doc -> BSONValue? in
```
jsflax: [nit] Use `compactMap` here. This will save us the step of having to do the `guard` later on.

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/DataSynchronizer.swift
```diff
@@ -507,8 +551,57 @@ public class DataSynchronizer: NetworkStateDelegate, FatalErrorListener {
...
+            return doc["_id"]
+        }
+
+        let result = try localColl.deleteMany(filter, options: options)
+
+        for documentId in idsToDelete {
```
jsflax: [nit] Why don't we map these into an array instead of a classic for/loop?

PR #117, opened 2018-12-13
STITCH-2220 Implement recovery logic in DataSynchronizer for CRUD

adamchel:
This PR adds the functionality for recovering namespaces that had interrupted CRUD operations. Once L2R/R2L logic is complete, there will be another PR for STITCH-2291 which will add the undo collection logic in conflict resolution functions, as well as add assertions that the undo collection is empty after any DataSynchronizer CRUD operation. That PR should be significantly smaller since all of the main recovery logic is implemented here.

I added various drive-by fixes, including a refactor of `InstanceSynchronization`. I'm not completely happy with how it's structured now, since there is a lot of code that can be cleaned up, but for now it works fine and it's relatively similar to what Android is doing. We can refactor this when we focus on cleanup tasks next week.


Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/DataSynchronizer.swift
```diff
@@ -416,6 +550,7 @@ public class DataSynchronizer: NetworkStateDelegate, FatalErrorListener {
     }
 
 
+    typealias DeferredBlock = (() -> Void)
```
jsflax: [nit] Not sure this type is complex enough to warrant a typealias

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/DataSynchronizer.swift
```diff
@@ -906,6 +1116,12 @@ public class DataSynchronizer: NetworkStateDelegate, FatalErrorListener {
         }
     }
 
+    // MARK: Utilities
+
+    private static func documentIdFilter(forId id: BSONValue) -> Document {
```
jsflax: [nit] I intentionally haven't added this method. It's actually more verbose than just writing out the document– would be nice to take it back out.

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/DataSynchronizer.swift
adamchel: I did it because `let deferredEventEmittingBlock: (() -> Void)? = nil` is a little weird to read but I could change it back

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/DataSynchronizer.swift
adamchel: fair enough, getting rid of it.

PR #119, opened 2018-12-17
STITCH-2291 Recovery logic in sync conflict handling functions

adamchel:
This PR adds the remaining undo collection logic in `DataSynchronizer.swift` for correct recovery in all failure cases, and also tests to make sure that undo documents are not leaked in any normal CRUD scenario. 

I also drove-by the remaining deferred event emitting blocks, and made the unit tests compile again.


Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/DataSynchronizer.swift
```diff
@@ -1140,13 +1140,19 @@ public class DataSynchronizer: NetworkStateDelegate, FatalErrorListener {
...
+
         nsConfig.nsLock.writeLock()
         defer { nsConfig.nsLock.unlock() }
 
         nsConfig[documentId] = nil
         try self.localCollection(for: namespace,
                                  withType: Document.self).deleteOne(["_id": documentId])
-        triggerListening(to: namespace)
+
+        deferredEventEmittingBlock = {
```
jsflax: [q] What was the point of this change?

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/DataSynchronizer.swift
```diff
@@ -1174,19 +1183,32 @@ public class DataSynchronizer: NetworkStateDelegate, FatalErrorListener {
...
+        }
 
         let event = ChangeEvent<Document>.changeEventForLocalReplace(namespace: namespace,
                                                                      documentId: documentId,
                                                                      document: docForStorage,
                                                                      writePending: false)
-        self.emitEvent(documentId: documentId,
-                       event: event)
+
+        deferredEventEmittingBlock = {
```
jsflax: [q] Same as above, any reason this is deferred?

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/DataSynchronizer.swift
adamchel: parity with android

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/DataSynchronizer.swift
adamchel: parity with android

jsflax: I feel like this is such a small detail that parity doesn't need to be maintained.

adamchel: I don't think it's a small detail, this is the only way to emit the event after the lock is unlocked, which is what happens in android

jsflax: Ah I see. Makes sense.

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/DataSynchronizer.swift
```diff
@@ -1140,13 +1140,19 @@ public class DataSynchronizer: NetworkStateDelegate, FatalErrorListener {
...
         nsConfig.nsLock.writeLock()
         defer { nsConfig.nsLock.unlock() }
 
         nsConfig[documentId] = nil
         try self.localCollection(for: namespace,
                                  withType: Document.self).deleteOne(["_id": documentId])
-        triggerListening(to: namespace)
+
+        deferredEventEmittingBlock = {
+            self.triggerListening(to: namespace)
```
dkaminsky: this logic is not conditional on anything -- why not just defer this block directly? rather than defer the reference above then change the value

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/DataSynchronizer.swift
```diff
@@ -1174,19 +1183,32 @@ public class DataSynchronizer: NetworkStateDelegate, FatalErrorListener {
...
 
         let event = ChangeEvent<Document>.changeEventForLocalReplace(namespace: namespace,
                                                                      documentId: documentId,
                                                                      document: docForStorage,
                                                                      writePending: false)
-        self.emitEvent(documentId: documentId,
-                       event: event)
+
+        deferredEventEmittingBlock = {
+            self.emitEvent(documentId: documentId,
```
dkaminsky: ditto re: why not just directly deferred

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/DataSynchronizer.swift
```diff
@@ -1248,7 +1280,13 @@ public class DataSynchronizer: NetworkStateDelegate, FatalErrorListener {
...
             atVersion: atVersion,
             changeEvent: event);
-        emitEvent(documentId: documentId, event: event)
+        if documentBeforeUpdate != nil {
+            try undoCollection.deleteOne([idField: documentId])
+        }
+
+        deferredEventEmittingBlock = {
+            self.emitEvent(documentId: documentId, event: event)
+        }
```
dkaminsky: ditto

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/DataSynchronizer.swift
```diff
@@ -1274,17 +1315,24 @@ public class DataSynchronizer: NetworkStateDelegate, FatalErrorListener {
...
+        try undoCollection.insertOne(documentToDelete)
+        try localCollection.deleteOne([idField: documentId])
         try desyncDocumentFromRemote(namespace: namespace, documentId: documentId)
-        emitEvent(documentId: documentId,
-                  event: ChangeEvent<Document>.changeEventForLocalDelete(namespace: namespace,
-                                                                         documentId: documentId,
-                                                                         writePending: false))
+        try undoCollection.deleteOne([idField: documentId])
+
+        deferredEventEmittingBlock = {
```
dkaminsky: ditto

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/DataSynchronizer.swift
```diff
@@ -1294,19 +1342,34 @@ public class DataSynchronizer: NetworkStateDelegate, FatalErrorListener {
...
         try localCollection.deleteOne(["_id": documentId])
         let event = ChangeEvent<Document>.changeEventForLocalDelete(namespace: namespace, documentId: documentId, writePending: true);
         try config.setSomePendingWrites(atTime: logicalT, atVersion: atVersion, changeEvent: event)
-        emitEvent(documentId: documentId, event: event);
+
+        if documentToDelete != nil {
+            try undoCollection.deleteOne([idField: documentId])
+        }
+
+        deferredEventEmittingBlock = {
```
dkaminsky: ditto

Darwin/Services/StitchRemoteMongoDBService/StitchRemoteMongoDBServiceTests/SyncIntTests.swift
```diff
@@ -1936,8 +2010,12 @@ class SyncIntTests: BaseStitchIntTestCocoaTouch {
 
         XCTAssertNil(coll.findOne(doc1Filter))
         XCTAssertNil(remoteColl.findOne(doc1Filter))
+
+        coll.verifyUndoCollectionEmpty()
     }
 
+    // MARK: Utilities
```
dkaminsky: [q] unfamiliar with this notation, can you explain?

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/DataSynchronizer.swift
adamchel: There is the unlikely possibility that `deleteOne` might fail (in which case we have a larger problem), so we don't necessarily want to trigger listening to the namespace in that case

Darwin/Services/StitchRemoteMongoDBService/StitchRemoteMongoDBServiceTests/SyncIntTests.swift
adamchel: Just a marker for XCode so that you can quickly jump to this section of the file from the UI:

image

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/DataSynchronizer.swift
adamchel: See response in slack. Copied here: "the reason we do that deferred pattern is because defers are done in FIFO order. the lock is the first thing we explicitly defer, but there are certain things we want to do after the lock is unlocked, such as triggering listening to a namespace or emitting events"

adamchel:
@dkaminsky PTAL

PR #122, opened 2018-12-20
STITCH-2268 add lint task to evergreen

adamchel:
Adds a working lint task to evergreen.

This also drive-bys most of the necessary changes for STITCH-2266, but until the Evergreen team fixes the xcodebuild segfault problem on the hosts, STITCH-2266 can't be completed.

### DIFF
--- a/.evg.yml
+++ b/.evg.yml
@@ -7,10 +7,11 @@ functions:
     - command: shell.exec
       params:
         script: |
-          git clone https://github.com/realm/SwiftLint
-          cd SwiftLint
-          git submodule update --init --recursive
-          make install
+          mkdir -p SwiftLint
+          pushd SwiftLint
+          curl -o swiftlint.zip -L https://github.com/realm/SwiftLint/releases/download/0.29.1/portable_swiftlint.zip
+          unzip swiftlint.zip
+          popd
   "setup_mongod":
     - command: shell.exec
       params:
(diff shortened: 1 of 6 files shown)


.evg.yml
```diff
@@ -7,10 +7,11 @@ functions:
...
         script: |
-          git clone https://github.com/realm/SwiftLint
-          cd SwiftLint
-          git submodule update --init --recursive
-          make install
+          mkdir -p SwiftLint
+          pushd SwiftLint
+          curl -o swiftlint.zip -L https://github.com/realm/SwiftLint/releases/download/0.29.1/portable_swiftlint.zip
+          unzip swiftlint.zip
+          popd
```
dkaminsky: generally fine but what's the behavior if curl or unzip fails? does the script abort or continue?

.evg.yml
```diff
@@ -94,8 +96,11 @@ tasks:
         params: 
           script: |
             set -e
-            cd stitch-ios-sdk/StitchCore
-            swiftlint
+            pushd SwiftLint
+            export PATH=`pwd`:$PATH
```
dkaminsky: small warning that paths with whitespace may get messed up by this, consider 2-step process

.evg.yml
adamchel: I'll add a `set -e` so the script aborts

.evg.yml
adamchel: We use this pattern throughout `.evg.yml` because we know the paths won't have whitespace, so I'm going to leave as is.

PR #123, opened 2018-12-24
STITCH-2298 Swift SDK: Todo app for mobile sync

jsflax:
I would recommend checking out the branch and viewing the files that way– the RWLock changes make the diff a bit difficult to read.

- Change RWLock to use GCD
- Fix the way updateDescription is calculated for replaces
- Sanitize documents before updateDescription is calculated
- Add retry logic for ChangeStreamListeners
- Add ToDo App


Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/DataSynchronizer.swift
```diff
@@ -1134,24 +1125,20 @@ public class DataSynchronizer: NetworkStateDelegate, FatalErrorListener {
      * @param namespace  the namespace to put the document in.
      * @param documentId the _id of the document.
      */
-    public func desyncDocumentFromRemote(namespace: MongoNamespace,
+    internal func desyncDocumentFromRemote(nsConfig: NamespaceSynchronization,
                                          documentId: BSONValue) throws {
```
adamchel: [nit] vertical parameter alignment

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/DataSynchronizer.swift
```diff
@@ -1175,17 +1162,17 @@ public class DataSynchronizer: NetworkStateDelegate, FatalErrorListener {
         var deferredEventEmittingBlock: DeferredBlock? = nil
```
adamchel: [nit] with this new locking scheme, we don't need the deferred event emitting block. Since we're not deferring to unlock the lock, we can just emit the event at the end of the function

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/DataSynchronizer.swift
```diff
@@ -1194,16 +1181,19 @@ public class DataSynchronizer: NetworkStateDelegate, FatalErrorListener {
...
 
-        let event = ChangeEvent<Document>.changeEventForLocalReplace(namespace: namespace,
-                                                                     documentId: documentId,
-                                                                     document: docForStorage,
-                                                                     writePending: false)
+        let event = ChangeEvent<Document>.changeEventForLocalReplace(
+            namespace: namespace,
+            documentId: documentId,
+            document: docForStorage,
+            updateDescription: documentBeforeUpdate?.diff(otherDocument: docForStorage),
```
adamchel: [minor] this goes back to my comment earlier, but a local replace event should not have an updateDescription. A replace is a full document replace, so there is no need for an updateDescription.

Why did this get added?

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/DataSynchronizer.swift
```diff
@@ -1231,10 +1221,8 @@ public class DataSynchronizer: NetworkStateDelegate, FatalErrorListener {
         var deferredEventEmittingBlock: DeferredBlock? = nil
```
adamchel: [nit] with this new locking scheme, we don't need the deferred event emitting block. Since we're not deferring to unlock the lock, we can just emit the event at the end of the function

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/DataSynchronizer.swift
```diff
@@ -1296,35 +1286,32 @@ public class DataSynchronizer: NetworkStateDelegate, FatalErrorListener {
...
-        namespace: MongoNamespace,
-        documentId: BSONValue
-        ) throws {
-        guard let lock = syncConfig[namespace]?.nsLock else {
+    private func deleteOneFromRemote(namespace: MongoNamespace, documentId: BSONValue) throws {
+        guard let nsConfig = syncConfig[namespace] else {
             return
         }
 
         var deferredEventEmittingBlock: DeferredBlock? = nil
```
adamchel: [nit] with this new locking scheme, we don't need the deferred event emitting block. Since we're not deferring to unlock the lock, we can just emit the event at the end of the function

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/DataSynchronizer.swift
```diff
@@ -1345,12 +1332,13 @@ public class DataSynchronizer: NetworkStateDelegate, FatalErrorListener {
...
```
adamchel: [nit] with this new locking scheme, we don't need the deferred event emitting block. Since we're not deferring to unlock the lock, we can just emit the event at the end of the function

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/DataSynchronizer.swift
```diff
@@ -1376,57 +1364,56 @@ public class DataSynchronizer: NetworkStateDelegate, FatalErrorListener {
      Starts data synchronization in a background thread.
      */
     public func start() {
-        syncLock.writeLock()
-        defer { syncLock.unlock(for: .writing) }
-        if (!self.isConfigured) {
-            return
-        }
+        try? syncLock.write {
```
adamchel: [minor] the `try?` seems smelly to me. makes it seem like we're just squashing any errors that happen. The old implementation caught any errors that happened within this block, so if you catch all the errors here then the `lock.write` shouldn't be re-throwing

if you don't want to solve it here, add a TODO for the error handling refactor ticket

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/DataSynchronizer.swift
```diff
@@ -1538,11 +1530,11 @@ public class DataSynchronizer: NetworkStateDelegate, FatalErrorListener {
             throw StitchError.clientError(
                 withClientErrorCode: .couldNotLoadSyncInfo)
         }
-        lock.readLock()
-        defer { lock.unlock(for: .reading) }
+        return try lock.read {
 
         return try localCollection(for: namespace, withType: Document.self)
```
adamchel: [nit] indent

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/DataSynchronizer.swift
```diff
@@ -1570,10 +1562,10 @@ public class DataSynchronizer: NetworkStateDelegate, FatalErrorListener {
             throw StitchError.clientError(
                 withClientErrorCode: .couldNotLoadSyncInfo)
         }
-        lock.readLock()
-        defer { lock.unlock(for: .reading) }
+        return try lock.read {
 
         return try localCollection(for: namespace).find(filter, options: options)
```
adamchel: [nit] indent

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/DataSynchronizer.swift
```diff
@@ -1592,12 +1584,12 @@ public class DataSynchronizer: NetworkStateDelegate, FatalErrorListener {
             throw StitchError.clientError(
                 withClientErrorCode: .couldNotLoadSyncInfo)
         }
-        lock.readLock()
-        defer { lock.unlock(for: .reading) }
+        return try lock.read {
 
         return try localCollection(for: namespace, withType: Document.self).aggregate(
```
adamchel: [nit] indent

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/DataSynchronizer.swift
```diff
@@ -1621,26 +1613,28 @@ public class DataSynchronizer: NetworkStateDelegate, FatalErrorListener {
         var deferredEventEmittingBlock: DeferredBlock? = nil
```
adamchel: [nit] with this new locking scheme, we don't need the deferred event emitting block. Since we're not deferring to unlock the lock, we can just trigger listening and emit the event at the end of the function outside of the `nsLock.write` block

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/DataSynchronizer.swift
```diff
@@ -1663,8 +1657,7 @@ public class DataSynchronizer: NetworkStateDelegate, FatalErrorListener {
         defer { deferredEventEmittingBlock?() }
```
adamchel: [nit] with this new locking scheme, we don't need the deferred event emitting block. Since we're not deferring to unlock the lock, we can just trigger listening and emit the event at the end of the function outside of the `lock.write` block

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/DataSynchronizer.swift
```diff
@@ -1714,8 +1708,7 @@ public class DataSynchronizer: NetworkStateDelegate, FatalErrorListener {
...
```
adamchel: [nit] with this new locking scheme, we don't need the deferred event emitting block. Since we're not deferring to unlock the lock, we can just trigger listening and emit the event at the end of the function outside of the `lock.write` block

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/DataSynchronizer.swift
```diff
@@ -1779,8 +1773,7 @@ public class DataSynchronizer: NetworkStateDelegate, FatalErrorListener {
...
```
adamchel: [nit] with this new locking scheme, we don't need the deferred event emitting block. Since we're not deferring to unlock the lock, we can just trigger listening and emit the event at the end of the function outside of the `lock.write` block

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/DataSynchronizer.swift
```diff
@@ -1847,101 +1841,101 @@ public class DataSynchronizer: NetworkStateDelegate, FatalErrorListener {
...
```
adamchel: [nit] with this new locking scheme, we don't need the deferred event emitting block. Since we're not deferring to unlock the lock, we can just trigger listening and emit the event at the end of the function outside of the `lock.write` block

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/DataSynchronizer.swift
```diff
@@ -1968,8 +1962,7 @@ public class DataSynchronizer: NetworkStateDelegate, FatalErrorListener {
...
```
adamchel: [nit] with this new locking scheme, we don't need the deferred event emitting block. Since we're not deferring to unlock the lock, we can just trigger listening and emit the event at the end of the function outside of the `lock.write` block

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/DataSynchronizer.swift
```diff
@@ -2086,15 +2080,15 @@ public class DataSynchronizer: NetworkStateDelegate, FatalErrorListener {
      - parameter event:      the change event.
      */
     private func emitEvent(documentId: BSONValue, event: ChangeEvent<Document>) {
-        listenersLock.readLock()
-        defer { listenersLock.unlock(for: .reading) }
+//        listenersLock.write {
```
adamchel: [q] this shouldn't be commented out right?

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/BSONFixes.swift
```diff
@@ -0,0 +1,39 @@
+import MongoSwift
+
+func validateBSONTypes(_ lhs: BSONValue, _ rhs: BSONValue) {
```
adamchel: [nit] maybe add TODOs in this file referring to the `SWIFT-` tickets that they are waiting on

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/UpdateDescription.swift
```diff
@@ -8,6 +8,16 @@ public class UpdateDescription: Codable {
     public let updatedFields: Document
     public let removedFields: [String]
 
+    internal init?(from document: Document) {
```
adamchel: [nit] this doesn't seem to have any callers, so remove.

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/UpdateDescription.swift
```diff
@@ -8,6 +8,16 @@ public class UpdateDescription: Codable {
...
+        guard let updatedFields = document["$set"] as? Document,
```
adamchel: if this initializer is actually supposed to exist, this is incorrect. an update description might only have one of `$set` or `$unset`

Darwin/Examples/ToDoSync/ToDoSync/AppDelegate.swift
```diff
@@ -0,0 +1,38 @@
+import UIKit
+import StitchCore
+import StitchRemoteMongoDBService
+
+private let todoListsDatabase = "todo"
+private let todoItemsCollection = "items"
+private let todoListsCollection = "lists"
+private let todoIndexSwapsCollection = "index_swaps"
+
+let stitch = try! Stitch.initializeAppClient(withClientAppID: "todo-cqenc")
```
adamchel: [nit q] are we fine hardcoding this stitch client app id? we do it in android too but it means anyone who looks at this repo and can see & edit the todo items in this app

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/NamespaceChangeStreamDelegate.swift
```diff
@@ -45,34 +46,57 @@ class NamespaceChangeStreamDelegate: SSEStreamDelegate, NetworkStateDelegate {
...
+                            sleep(retrySleepSeconds)
+                        }
+                    } catch {
+                        self.logger.e("NamespaceChangeStreamRunner::run error happened while opening stream: \(error)");
+                        return
+                    }
+                } else {
+                    self.holdingSemaphore.wait()
+                }
+            } while self?.networkMonitor.state == .connected &&
```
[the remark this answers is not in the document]
jsflax: The loop would end if the task is closed/cancelled, which is a natural occurrence in this loop. Another option would be to make the work item `weak`. This is intrinsically the same pattern as Android– Android just does it by blocking by reading from the open stream.

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/NamespaceChangeStreamDelegate.swift
```diff
@@ -45,34 +46,57 @@ class NamespaceChangeStreamDelegate: SSEStreamDelegate, NetworkStateDelegate {
...
+                        let isOpening = try self.openStream()
+                        if !isOpening {
+                            sleep(retrySleepSeconds)
+                        }
+                    } catch {
+                        self.logger.e("NamespaceChangeStreamRunner::run error happened while opening stream: \(error)");
+                        return
+                    }
+                } else {
+                    self.holdingSemaphore.wait()
```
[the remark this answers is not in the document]
jsflax: I can add a comment, but this actually isn't terribly different than Android.

Core/StitchCoreSDK/Sources/StitchCoreSDK/Internal/Net/FoundationHTTPTransport.swift
```diff
@@ -69,23 +69,30 @@ public final class FoundationHTTPTransport: Transport {
...
     public func stream(request: Request, delegate: SSEStreamDelegate? = nil) throws -> RawSSEStream {
         guard let url = URL(string: request.url) else {
             throw StitchError.clientError(withClientErrorCode: .missingURL)
         }
 
         let sessionConfig = URLSessionConfiguration.default
-        sessionConfig.timeoutIntervalForRequest = 60
-        sessionConfig.timeoutIntervalForResource = 60
-        sessionConfig.requestCachePolicy = .reloadIgnoringLocalCacheData
+        sessionConfig.timeoutIntervalForRequest = TimeInterval(INT_MAX)
```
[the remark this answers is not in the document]
jsflax: I will have to do internal testing to see how this affects it. I'm going by what other SSE libraries did– I think the timeout here might be a regardless of heartbeats.

Core/StitchCoreSDK/Sources/StitchCoreSDK/Internal/Net/RawSSEStream.swift
```diff
@@ -104,7 +108,7 @@ open class RawSSEStream {
      Process and dispatch the events in a given stream.
      */
     internal func dispatchEvents() {
-        while state == .open, let line = self.readLine() {
+        while state == .opening || state == .open, let line = self.readLine() {
```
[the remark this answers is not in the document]
jsflax: Seems like it can be changed back, yes.

Core/StitchCoreSDK/Sources/StitchCoreSDK/Internal/Net/FoundationHTTPTransport.swift
```diff
@@ -69,23 +69,30 @@ public final class FoundationHTTPTransport: Transport {
...
+        sessionConfig.timeoutIntervalForResource = TimeInterval(INT_MAX)
         let additionalheaders = [Headers.contentType.nonCanonical(): "text/event-stream",
                                  Headers.cacheControl.nonCanonical(): "no-cache",
                                  Headers.accept.nonCanonical(): "text/event-stream"]
         sessionConfig.httpAdditionalHeaders = additionalheaders
         let sseStream = FoundationHTTPSSEStream(delegate)
         let session = URLSession.init(configuration: sessionConfig,
                                       delegate: sseStream.dataDelegate,
-                                      delegateQueue: nil)
+                                      delegateQueue: opQueue)
```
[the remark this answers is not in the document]
jsflax: Yes.

Core/StitchCoreSDK/StitchCoreSDK.xcodeproj/project.pbxproj
```diff
@@ -1269,6 +1272,7 @@
 				DYLIB_COMPATIBILITY_VERSION = 1;
 				DYLIB_CURRENT_VERSION = 1;
 				DYLIB_INSTALL_NAME_BASE = "@rpath";
+				INFOPLIST_FILE = "/Users/jasonflax/Development/stitch-ios-sdk/Core/StitchCoreSDK/StitchCoreSDK.xcodeproj/StitchCoreSDK_Info.plist";
```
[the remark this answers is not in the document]
jsflax: Not sure how that got in there.

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/DataSynchronizer.swift
jsflax: [style]

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/DataSynchronizer.swift
jsflax: It seems very odd that remote updates that are "coalesced" into replaces lose the update information. This resulted in a weird UX for the app I was building.

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/DataSynchronizer.swift
jsflax: We can bubble up the errors here into the fatal error listener.

Darwin/Examples/ToDoSync/ToDoSync/AppDelegate.swift
jsflax: I don't see an issue with that at the moment. I agree, it's odd, but can be removed down the line.

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/NamespaceChangeStreamDelegate.swift
```diff
@@ -109,20 +134,19 @@ class NamespaceChangeStreamDelegate: SSEStreamDelegate, NetworkStateDelegate {
...
-        defer { eventQueueLock.unlock(for: .writing) }
-
         do {
-            guard let changeEvent: ChangeEvent<Document> = try event.decodeStitchSSE(),
-                let id = changeEvent.documentKey["_id"] else {
-                return
-            }
-
-            logger.d("event found: op=\(changeEvent.operationType) id=\(id)")
+            try eventQueueLock.write {
```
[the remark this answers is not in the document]
jsflax: `stop` calls the write lock itself.

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/DataSynchronizer.swift
```diff
@@ -310,30 +296,30 @@ public class DataSynchronizer: NetworkStateDelegate, FatalErrorListener {
...
+        }
 
-        self.syncConfig = try InstanceSynchronization(configDb: configDb, errorListener: self)
-        self.isConfigured = false
         self.stop()
     }
 
     func doSyncPass() throws -> Bool {
-        guard isConfigured, syncLock.tryLock(for: .reading) else {
+        guard isConfigured else {
```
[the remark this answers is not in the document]
jsflax: Because we're already holding the lock within a writeLock within `start`.

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/DataSynchronizer.swift
jsflax: Actually there are no errors here.

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/DataSynchronizer.swift
```diff
@@ -1059,9 +1045,14 @@ public class DataSynchronizer: NetworkStateDelegate, FatalErrorListener {
             }
         }
 
-        let acceptRemote = (remoteEvent.fullDocument == nil && resolvedDocument == nil)
-            || (remoteEvent.fullDocument != nil
-                && remoteEvent.fullDocument == resolvedDocument)
+        var sanitizedRemoteDocument: Document? = nil
+        if let remoteDocument = remoteEvent.fullDocument {
+            sanitizedRemoteDocument = DataSynchronizer.sanitizeDocument(remoteDocument)
```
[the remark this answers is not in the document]
jsflax: [nit] -> [q]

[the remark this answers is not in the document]
jsflax: I do not know. I can add a note for myself externally to check this later on, but Android as a whole is out of scope for this ticket.

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/DataSynchronizer.swift
```diff
@@ -162,20 +160,14 @@ public class DataSynchronizer: NetworkStateDelegate, FatalErrorListener {
      */
     private func recover(recoveryStarted: DispatchSemaphore) throws {
         let nsConfigs = self.syncConfig.map { $0 }
-        nsConfigs.forEach { namespaceSynchronization in
```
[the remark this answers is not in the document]
jsflax: 👍

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/ChangeEvent.swift
```diff
@@ -179,14 +179,15 @@ public struct ChangeEvent<DocumentT: Codable>: Codable, Hashable {
     static func changeEventForLocalReplace(namespace: MongoNamespace,
                                            documentId: BSONValue,
                                            document: Document,
+                                           updateDescription: UpdateDescription? = nil,
```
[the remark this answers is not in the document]
adamchel: Replace events don't contain update descriptions per the MongoDB change event spec
https://docs.mongodb.com/manual/reference/change-events/#replace-event

When you say the update description contains vital information, do you mean vital to the end user or vital for the synchronization to happen properly? If it's the latter, then I think our coalescence algorithm is incorrect. What kind of information is "vital"? 

Seems like a user should be able to find everything they need to know from the fullDocument. If they need to see the update description, then maybe we shouldn't be coalescing the events in the first place. thoughts @edaniels @dkaminsky ?

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/CoreDocumentSynchronizationConfig.swift
```diff
@@ -151,52 +152,36 @@ internal class CoreDocumentSynchronization: Hashable {
     /// The most recent pending change event
     var uncommittedChangeEvent: ChangeEvent<Document>? {
         get {
```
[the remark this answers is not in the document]
adamchel: let's add the precondition checks then. this is all internal stuff so I think it's fine for us to do that, and will prevent us from incorrectly calling these without locks in the future.

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/ChangeEvent.swift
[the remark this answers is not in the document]
jsflax: For my use case, I'm updating the `index` of a `TodoItem`. This is to make sure that all TodoLists are ordered the same. Kind of a handy feature, and the todo apps I use definitely sync ordering.

There are specific actions I want to take in the event that an `index` changes– namely, to change the ordering of the todoList. For this purpose, I have an `index_swap` table– a single document keyed on `owner_id` that tells me which two elements to swap on the list. However, I'm still going to receive the update for the index in the `items` collection. While I want to update the internal list with the new information, I do _not_ want to update the tableView, as it's being updated from the `index_swap`. So I check for whether or not the `index` field has been updated and if it has, I do not update the tableView. Now, I could discern this via running my own diff, but that seems unnecessary.

This seems like it would be a fairly common use case– not for todo lists per se, but most updates are going to be coalesced to replaces. Which means most update descriptions are going to be written off.

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/DataSynchronizer.swift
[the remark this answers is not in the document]
adamchel: if we're holding a write lock within `start`, then how is `doSyncPass` able to call other functions that take the `syncLock`, like `triggerListening`, and how would users be able to call `configure`?

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/DataSynchronizer.swift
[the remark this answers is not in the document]
adamchel: Android is out of scope, but if we uncover bugs here that exist in Android, then we can't just ignore them. This is where I'm coming from with pushing for the C++ core module

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/DataSynchronizer.swift
adamchel: My comments about getting rid of the deferred event emitting block are still unaddressed. We can leave as is if we think we might move away from this locking mechanism eventually but if we're sticking with the dispatch approach then you can move anything inside the deferred event emitting block at the end to just occur outside the locked block

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/ChangeEvent.swift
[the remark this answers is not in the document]
adamchel: Hmm, so I personally think it will probably be a best practice for users to do their own diff since change events my come from various sources outside of their own app, that won't be guaranteed to be an update (or vice versa).

though I get where you're coming from, and it is somewhat jarring for a simple app where the user knows all the events they're doing. Perhaps it would make sense for coalescence to be a configurable setting? 

I get that having the update description in the replace will be useful, but then we are diverging from something standard that MongoDB does, and it will set up an expectation from users that all replace events will have an updatedescription, when that is not the case. Replace events coming from the Stitch server will have no update description.

@dkaminsky @edaniels what are your thoughts on making change event coalescence a toggle-able setting?

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/DataSynchronizer.swift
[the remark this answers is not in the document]
jsflax: The write lock is only held until the thread is dispatched

jsflax:
@adamchel PTAL

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/DataSynchronizer.swift
```diff
@@ -1629,13 +1618,16 @@ public class DataSynchronizer: NetworkStateDelegate, FatalErrorListener {
             namespace: namespace,
             document: docForStorage,
             writePending: true)
-        let config = try nsConfig.sync(id: result.insertedId!)
-        try config.setSomePendingWrites(atTime: logicalT, changeEvent: event)
 
-        deferredEventEmittingBlock = {
+        try nsConfig.nsLock.write {
```
adamchel: [minor q] why is there a gap in what is locked here. Everything before the trigger listening call should fall under the same critical section.

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/DataSynchronizer.swift
```diff
@@ -1648,43 +1640,45 @@ public class DataSynchronizer: NetworkStateDelegate, FatalErrorListener {
...
+                let document = docsForStorage[kv.key]
+                let event = ChangeEvent<Document>.changeEventForLocalInsert(namespace: namespace,
+                                                                            document: document,
+                                                                            writePending: true)
+                let config = try nsConfig.sync(id: documentId)
+                try config.setSomePendingWrites(atTime: logicalT, changeEvent: event)
+                return { self.emitEvent(documentId: documentId, event: event) }
+            })
+
             eventEmitters.forEach({$0()})
```
adamchel: [nit] This should be moved to happen after the `triggerListening` call, since right now the events are being emitted inside the lock, which isn't necessary

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/DataSynchronizer.swift
```diff
@@ -1769,52 +1768,56 @@ public class DataSynchronizer: NetworkStateDelegate, FatalErrorListener {
...
+                        try undoColl.deleteOne([idField: documentId])
+                    }
+                    return
+                }
 
-                try desync(ids: [docConfig.documentId.value], in: docConfig.namespace)
+                try docConfig.setSomePendingWrites(atTime: logicalT, changeEvent: event)
                 try undoColl.deleteOne([idField: documentId])
-                return nil
+                self.emitEvent(documentId: documentId, event: event)
```
adamchel: [nit] this can be moved outside of the critical section

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/DataSynchronizer.swift
```diff
@@ -1924,18 +1924,21 @@ public class DataSynchronizer: NetworkStateDelegate, FatalErrorListener {
                 try undoColl.deleteOne([idField: documentIdBeforeUpdate])
             }
 
-            deferredEventEmittingBlock = {
-                if triggerNamespace {
-                    self.triggerListening(to: namespace)
-                }
-                self.emitEvent(documentId: documentId, event: event);
-            }
+            self.emitEvent(documentId: documentId, event: event);
```
adamchel: [nit] this can be moved outside the critical section after the triggerListening call

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/DataSynchronizer.swift
adamchel: and do this for all `emitEvent` calls

PR #128, opened 2019-01-07
STITCH-2248 Swift SDK: Split [dataSync] errors between recoverable and irrecovera…

jsflax:
…ble errors

### DIFF
--- a/Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/DataSynchronizerError.swift
+++ b/Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/DataSynchronizerError.swift
@@ -1,9 +1,65 @@
-public class DataSynchronizerError: Error, CustomStringConvertible {
-    public var description: String
+/// A set of errors that can occur during the synchronization process.
+public enum DataSynchronizerError: Error, CustomStringConvertible {
+    /**
+        Any error related to decoding a document
+        This will likely be due to a malformed sync version
+        or an unsupported change event.
+    */
+    case decodingError(_ message: String)
+    /**
+        An unsupported sync protocol version has been
+        supplied with a document. The error message
+        contains the faulty version.
+    */
+    case unsupportedProtocolVersion(_ message: String)
+    /**
+        Any error related to trying to operate
+        on a document that does not exist.
+        This can occur if a change event is sent to us
+        with missing data, or if a a local document
+        is deleted during a replace or update.
+     */
+    case documentDoesNotExist(_ message: String)
+    /**
+        Any unexpected mongoDB error. Certain errors
+        during the syncing process are expected and will
+        create conflicts (e.g., inserting on the same _id)
+        but errors outside of the scope of what we can handle
+        will be thrown here
+    */
+    case mongoDBError(_ message: String, _ error: Error)
+    /**
+        Any error that occurs during the resolution of
+        a conflict. Generally, these will be issues
+        that occur in the user's delegate class or
+        callback.
+    */
+    case resolutionError(_ error: Error)
+    /**
+        A special type of error related to irrecoverable
+        issues during the synchronization process. This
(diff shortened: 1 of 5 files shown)


Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/DataSynchronizerError.swift
```diff
@@ -1,9 +1,65 @@
...
+        This will likely be due to a malformed sync version
+        or an unsupported change event.
+    */
+    case decodingError(_ message: String)
+    /**
+        An unsupported sync protocol version has been
+        supplied with a document. The error message
+        contains the faulty version.
+    */
+    case unsupportedProtocolVersion(_ message: String)
```
dkaminsky: isn't this non-recoverable?

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/DataSynchronizerError.swift
```diff
@@ -1,9 +1,65 @@
...
+     */
+    case documentDoesNotExist(_ message: String)
+    /**
+        Any unexpected mongoDB error. Certain errors
+        during the syncing process are expected and will
+        create conflicts (e.g., inserting on the same _id)
+        but errors outside of the scope of what we can handle
+        will be thrown here
+    */
+    case mongoDBError(_ message: String, _ error: Error)
```
dkaminsky: does this capture the error category from newer versions of mongo (e.g. ability to see if this was a transient network error, etc)?

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/DataSynchronizerError.swift
jsflax: No, a non-recoverable error would be one that forces the data synchronizer to shut down, e.g., not being able to interact with embedded mongo (for whatever reason).

An unsupported protocol version would, however, "pause" the document that is being sync'd. This would then be made available in the `pausedIds` field of the user-exposed `Sync` class. Users could then choose how to fix the error.

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/DataSynchronizerError.swift
jsflax: This captures whatever errors could be emitted by MobileMongo.

jsflax:
The user's error delegate would receive these errors:

```swift
func on(error: DataSynchronizerError, forDocumentId documentId: BSONValue?) {
        switch error {
        case .decodingError(_):
            // if I receive this error, the message will contain information
            // telling me that I've messed with a documents version info to the point
            // that it can no longer be read or decoded.
            // So if that happens, I could fix the error, and then unpause
            // the document.
            break
        case .unsupportedProtocolVersion(_):
            // if I receive this error, the message will contain information
            // telling me that I'm using an unsupported protocol version.
            break
        case .documentDoesNotExist(_):
            // if I receive this error, chances are I've done something in my
            // own application that is racy.
            break
        case .mongoDBError(_, _):
            // if I receive this error, there will be more information within the
            // secondary error as to what I've done
            break
        case .resolutionError(_):
            // if I received this error, I would look to my conflict
            // resolver to see what error could be surfacing here
            break
        case .fatalError(let msg):
            // This error tells me that we've met with a terrible fate.
            // The user is likely trying to use the application on a system
            // that embedded mongo does not support.
            fatalError(msg.localizedDescription)
        }
    }
```

It's generally so the user could debug. Documents that have been paused will not be synced on. So a user would have to fix their issues, and then unpause the document manually, which we expose methods for.

PR #133, opened 2019-01-23
STITCH-2443 Swift SDK: Upgrade MongoSwift and MongoMobile to 0.0.8/0.0.6

jsflax:


Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/Internal/CoreDocumentSynchronizationConfig.swift
```diff
@@ -86,10 +89,15 @@ final class CoreDocumentSynchronization: Codable, Hashable {
     var isStale: Bool {
         get {
             return docLock.read {
-                var filter = docConfigFilter(forNamespace: namespace, withDocumentId: documentId.bsonValue)
+                var filter = docConfigFilter(forNamespace: namespace, withDocumentId: documentId).map {
```
adamchel: [task] can you add a quick comment explaining what's happening here? it seems like a contrived way of adding a field to a BSON document that I assume is because of a Swift driver change, but we should be clear about it

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/Internal/CoreDocumentSynchronizationConfig.swift
```diff
@@ -12,8 +12,11 @@ import StitchCoreSDK
  */
 internal func docConfigFilter(forNamespace namespace: MongoNamespace,
                               withDocumentId documentId: AnyBSONValue) -> Document {
-    return [CoreDocumentSynchronization.CodingKeys.namespace.rawValue:
-        try? BSONEncoder().encode(namespace),
+    guard let namespaceDoc = try? BSONEncoder().encode(namespace) else {
+        return [CoreDocumentSynchronization.CodingKeys.namespace.rawValue: BSONNull(),
+         CoreDocumentSynchronization.CodingKeys.documentId.rawValue: documentId.value]
+    }
+    return [CoreDocumentSynchronization.CodingKeys.namespace.rawValue: namespaceDoc,
```
adamchel: [style nit] could we get rid of the guard block and simplify this to `try? BSONEncoder().encode(namespace) ?? BSONNull()`?

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/Internal/CoreDocumentSynchronizationConfig.swift
```diff
@@ -299,7 +307,10 @@ final class CoreDocumentSynchronization: Codable, Hashable {
     }
 
     internal static func filter(forNamespace namespace: MongoNamespace) -> Document {
-        return [CodingKeys.namespace.rawValue: try? BSONEncoder().encode(namespace)]
+        guard let namespaceDoc = try? BSONEncoder().encode(namespace) else {
+            return [CodingKeys.namespace.rawValue: BSONNull()]
+        }
+        return [CodingKeys.namespace.rawValue: namespaceDoc]
```
adamchel: [style nit] could we get rid of the guard block and simplify this to `try? BSONEncoder().encode(namespace) ?? BSONNull()`?

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/Internal/NamespaceSynchronizationConfig.swift
```diff
@@ -13,7 +13,10 @@ import StitchCoreSDK
  */
 final class NamespaceSynchronization: Sequence, Codable {
     static func filter(namespace: MongoNamespace) -> Document {
-        return ["namespace": try? BSONEncoder().encode(namespace)]
+        guard let namespaceDoc = try? BSONEncoder().encode(namespace) else {
+            return ["namespace": BSONNull()]
+        }
+        return ["namespace": namespaceDoc]
```
adamchel: [style nit] could we get rid of the guard block and simplify this to `try? BSONEncoder().encode(namespace) ?? BSONNull()`?

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/Internal/ThreadSafeMongoMobile.swift
```diff
@@ -134,3 +134,29 @@ class ThreadSafeMongoCollection<T: Codable>: Codable {
         return try underlyingCollection().deleteMany(filter, options: options)
     }
 }
+
+extension ThreadSafeMongoCollection where T == Document {
+    @discardableResult
+    func insertOne(_ value: inout T) throws -> InsertOneResult? {
```
adamchel: [q] why did we do this? I know we're the only consumer of this but making these inout seems smelly and the behavior of modifying the argument seems a little odd to me, especially if the Swift driver isn't doing it. is there another way to do this?

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/Internal/CoreDocumentSynchronizationConfig.swift
jsflax: No, we cannot. This applies to all of your comments. The compiler doesn't like it.

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/Internal/ThreadSafeMongoMobile.swift
jsflax: This is the spec-based behaviour the driver used to have, that it has since removed. We rely heavily on this behaviour. The other way to do this would be to do this inline.

PR #137, opened 2019-01-25
Fix/demo

jsflax:


Core/Services/StitchCoreLocalMongoDBService/Sources/StitchCoreLocalMongoDBService/CoreLocalMongoDBService.swift
```diff
@@ -28,33 +29,35 @@ public final class CoreLocalMongoDBService {
...
-        }
+        return try cLock.write {
+            let key = "\(Thread.current.hash)_\(key)"
+            if let client = _localInstances[key] {
+                return client
+            }
 
-        objc_sync_enter(self)
-        defer { objc_sync_exit(self) }
+            objc_sync_enter(self)
```
adamchel: [q] If we're using the `cLock`, then is the `objc_sync_enter(self)` no longer necessary?

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/Internal/DataSynchronizer.swift
```diff
@@ -782,6 +782,7 @@ public class DataSynchronizer: NetworkStateDelegate, FatalErrorListener {
                                         // shouldn't update because it would improperly update the version information.
                                         logger.i("t='\(logicalT)': syncLocalToRemote ns=\(nsConfig.namespace) documentId=\(docConfig.documentId) local change event "
                                             + "update description is empty for UPDATE; dropping the event")
+                                        try docConfig.setPendingWritesComplete(atVersion: localVersionInfo.nextVersion)
```
adamchel: [minor + task] so this contradicts the comment here, which says "we shouldn't update because it would improperly update the version information". I'm assuming that the comment is wrong and this is the bug you were talking about in Slack. Can you update the comment here to reflect the reality (that we still need to update the document and its version since this update could be generated by us and not updating version would get us stuck in a pending state).

Also you should file a ticket to fix this bug in Android as well

PR #143, opened 2019-02-19
STITCH-1992 Add watch to RemoteMongoCollection

adamchel:
This adds support for opening `watch` streams on a `RemoteMongoCollection`. The interface to do this is pretty different from JavaScript, as I tried to mimic Foundation's `URLSession` to create a `ChangeStreamSession`. This was a bit complicated by the fact that ChangeEvents are generic on document type, but I was able to come up with an interface that I think will be pretty clean.

Below is a usage sample that might help in your review:

```swift
// this is a delegate that might be used to react to change events with a normal document type
class MyChangeStreamDelegate: ChangeStreamDelegate {
    // this delegate interacts with normal BSON documents
    typealias DocumentT = Document

    func didReceive(event: ChangeEvent<Document>) {
        // react to event
    }

    func didReceive(streamError: Error) {
        // react to error
    }

    func didOpen() {
        // react to stream opening
    }

    func didClose() {
        // react to stream closing
    }
}

class MyCustomTypeChangeStreamDelegate: ChangeStreamDelegate {
    // this delegate interacts with BSON documents of a custom user-defined type
    typealias DocumentT = MyCustomType

    func didReceive(event: ChangeEvent<MyCustomType>) {
        // react to event
    }

    func didReceive(streamError: Error) {
        // react to error
    }

    func didOpen() {
        // react to stream opening
    }

    func didClose() {
        // react to stream closing
    }
}

func myWatchSample() {
    // self.coll is assumed to be a remote mongo collection acting on normal documents
    let myDelegate = MyChangeStreamDelegate.init()
    let stream1 = self.coll.watch(["some_string_id", ObjectId("some_hex_string")], myDelegate)

    // self.customTypeColl is assumed to be a remote mongo collection acting on custom type docs
    let myCustomDelegate = MyCustomTypeChangeStreamDelegate.init()
    let stream2 = self.customTypeColl.watch(["some_other_id"], myCustomDelegate)

    // this won't compile, because the delegate is incompatible with the collection type
    let streamWontCompile = self.coll.watch(["some_other_id"], myCustomDelegate)

    // delegates can be hot-swapped 
    let someNewDelegate = MyDelegate.init()
    stream1.delegate = someNewDelegate

    // streams can be closed by closing their session
    stream1.close()
    stream2.close()
}
```


Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/ChangeEvent.swift
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    public init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+
+        self.id = try container.decode(AnyBSONValue.self, forKey: .id)
+        self.operationType = OperationType.init(
+            rawValue: try container.decode(String.self, forKey: .operationType)
+        ) ?? .unknown
+        self.fullDocument = try container.decodeIfPresent(DocumentT.self, forKey: .fullDocument)
```
tkaye407: curiosity question, is there no difference between the following two statements: 

self.blah = try? container.decode(Document.self, forKey: .key)
self.blah = try container.decodeIfPresent(Document.self, forKey: .key)

In my multiuser I used the try? convention so just curious

Darwin/Services/StitchRemoteMongoDBService/StitchRemoteMongoDBService/ChangeStreamSession.swift
```diff
@@ -0,0 +1,54 @@
...
+     *
+     * - warning the delegate is held by the change stream as a weak reference. If the delegate becomes
+     *           deallocated, the stream will automatically close the next time an event is received. If you are
+     *           setting a new delegate, make sure you do so before the previous delegate is deallocated.
+     */
+    public var delegate: DelegateT? {
+        get {
+            return self.internalDelegate.delegate
+        }
+        set(newValue) {
```
tkaye407: what would happen if you set the delegate to nil?

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/ChangeEvent.swift
adamchel: Those are not quite the same. If for whatever reason, the string we're decoding is malformed, `container.decodeIfPresent` can still throw and you get useful error information. The `try?` will squash decoding and presence errors and always just return `nil` in the case of an error.

`decodeIfPresent` is preferred since you can distinguish between decoding errors (an error will be thrown) and key presence errors (the result will be `nil`), so it's only effectively the same if you are treating the lack of key presence and decoding errors the same way, which I believe we are doing in multi-user

Darwin/Services/StitchRemoteMongoDBService/StitchRemoteMongoDBService/ChangeStreamSession.swift
adamchel: the next time the stream got an event, it would close the stream. 

this is optional because the underlying stream holds a weak var.

check out https://blog.bobthedeveloper.io/swift-retention-cycle-in-closures-and-delegate-836c469ef128 for why we hold the delegate weakly

PR #144, opened 2019-02-19
Find one

tkaye407:
I believe I added findOne() everywhere it is supposed to be, but let me know if I missed somewhere. 
I added tests for sync in RemoteMongoClientIntTests, but I did not add any in SyncIntTests because that did not seem necessary as I believe those only test Insert / Update / Delete.

### DIFF
--- a/Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Internal/CoreRemoteMongoCollection.swift
+++ b/Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Internal/CoreRemoteMongoCollection.swift
@@ -105,11 +105,7 @@ public class CoreRemoteMongoCollection<T: Codable> {
      *   - filter: A `Document` that should match the query.
      *   - options: Optional `RemoteFindOptions` to use when executing the command.
      *
-     * - important: Invoking this method by itself does not perform any network requests. You must call one of the
-     *              methods on the resulting `CoreRemoteMongoReadOperation` instance to trigger the operation against
-     *              the database.
-     *
-     * - returns: A `CoreRemoteMongoReadOperation` that allows retrieval of the resulting documents.
+     * - returns: The resulting `Document` or nil if no such document exists
      */
     public func findOne(_ filter: Document = [:], options: RemoteFindOptions? = nil) throws -> T? {
         var args = baseOperationArgs
--- a/Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Internal/CoreRemoteMongoCollection.swift
+++ b/Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Internal/CoreRemoteMongoCollection.swift
@@ -120,9 +120,9 @@ public class CoreRemoteMongoCollection<T: Codable> {
             }
         }
 
-        return try self.service.callFunctionOptional(withName: "findOne",
-                                                     withArgs: [args],
-                                                     withRequestTimeout: nil)
+        return try self.service.callFunctionOptionalResult(withName: "findOne",
+                                                           withArgs: [args],
+                                                           withRequestTimeout: nil)
     }
 
     /**
--- a/Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Internal/CoreRemoteMongoCollection.swift
+++ b/Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Internal/CoreRemoteMongoCollection.swift
@@ -120,16 +120,9 @@ public class CoreRemoteMongoCollection<T: Codable> {
             }
         }
 
-        do {
-            return try self.service.callFunction(withName: "findOne",
-                                                 withArgs: [args],
-                                                 withRequestTimeout: nil)
-        } catch let err {
-            if String(describing: err).contains("decodingErr") {
-                return nil
-            }
-            throw err
-        }
+        return try self.service.callFunctionOptional(withName: "findOne",
+                                                     withArgs: [args],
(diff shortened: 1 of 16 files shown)


Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Internal/CoreRemoteMongoCollection.swift
```diff
@@ -99,6 +98,40 @@ public class CoreRemoteMongoCollection<T: Codable> {
...
+        args["query"] = filter
+        if let options = options {
+            if let projection = options.projection {
+                args[RemoteFindOptionsKeys.projection.rawValue] = projection
+            }
+            if let sort = options.sort {
+                args[RemoteFindOptionsKeys.sort.rawValue] = sort
+            }
+        }
+
```
tkaye407: Below is definitely not the prettiest way to go about this, but I was not sure how / if we want to change the implementation of CallFunction in the case of the request returning nil. Simply catching seemed like the least invasive way to go about this, but let me know if there is a better way to go about this

Darwin/Services/StitchHTTPService/StitchHTTPServiceTests/HTTPServiceClientIntTests.swift
```diff
@@ -129,7 +129,7 @@ class HTTPServiceClientIntTests: BaseStitchIntTestCocoaTouch {
 
             XCTAssertEqual("200 OK", response!.status)
             XCTAssertEqual(200, response!.statusCode)
-            XCTAssertTrue(300 <= response!.contentLength && response!.contentLength <= 500)
```
tkaye407: This is to make evergreen work, we should figure out exactly what is going on here

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Internal/CoreRemoteMongoCollection.swift
adamchel: We definitely don't want to be parsing error message strings for a common type of result. We should probably add another version of `doAuthenticatedRequest` that returns an optional T, and the compiler should automatically pick that one in `findOne`.

It would look something like

```swift
    public func doAuthenticatedRequest<T: Decodable>(_ stitchReq: StitchAuthRequest) throws -> T? {
        let response = try self.doAuthenticatedRequest(stitchReq)
        do {
            guard let responseBody = response.body,
                let responseString = String.init(data: responseBody, encoding: .utf8) else {
                    throw StitchError.serviceError(
                        withMessage: StitchErrorCodable.genericErrorMessage(withStatusCode: response.statusCode),
                        withServiceErrorCode: .unknown)
            }

            do {
                return try BSONDecoder().decode(T?.self, from: responseString)
            } catch let err {
                throw StitchError.requestError(withError: err, withRequestErrorCode: .decodingError)
            }
        }
    }
```

assuming that BSON undefined can be decoded to nil. I just checked and it compiles, but it should be tested. If that doesn't work, we should discuss the best approach with the Swift driver team.

Darwin/Services/StitchHTTPService/StitchHTTPServiceTests/HTTPServiceClientIntTests.swift
adamchel: this should be fixed now. merge master into sync/rebase, and uncomment this

Core/StitchCoreSDK/Sources/StitchCoreSDK/Auth/Internal/CoreStitchAuth+StitchAuthRequestClient.swift
```diff
@@ -44,6 +44,24 @@ extension CoreStitchAuth {
...
+        do {
+            guard let responseBody = response.body,
+                let responseString = String.init(data: responseBody, encoding: .utf8) else {
+                    throw StitchError.serviceError(
+                        withMessage: StitchErrorCodable.genericErrorMessage(withStatusCode: response.statusCode),
+                        withServiceErrorCode: .unknown)
+            }
+
+            do {
+                return try BSONDecoder().decode(T.self, from: responseString)
```
adamchel: [minor] I'm not sure this is quite what you want. I think you just want to get rid of the `do-catch` block and replace it with

```swift
try BSONDecoder().decode(T?.self, from: responseString)
```
If there actually is a decoding error, you want to propagate it.

Though it's possible that the Swift driver doesn't support decoding a BSON undefined/null into a `nil`. In that case, I guess we can leave this as is, but this method will mask all decoding errors by returning a `nil` result. This is probably fine for now, but ideally we want to be able to distinguish `nil` from an invalid response.

Darwin/Services/StitchRemoteMongoDBService/StitchRemoteMongoDBServiceTests/RemoteMongoClientIntTests.swift
```diff
@@ -246,6 +246,7 @@ class RemoteMongoClientIntTests: BaseStitchIntTestCocoaTouch {
     func testFindOne() {
         let coll = getTestColl()
         var exp = expectation(description: "should not find any documents in empty collection")
+        print("Calling FindOne()")
```
adamchel: [nit] remove debug print

adamchel:
And make sure when you squash and merge, you add the ticket number and a short description to the title so the commit has the ticket number.

e.g.
`STITCH-2525 add support for findOne`

Core/StitchCoreSDK/Sources/StitchCoreSDK/Auth/Internal/CoreStitchAuth+StitchAuthRequestClient.swift
tkaye407: Sadly, that does not work :(

PR #146, opened 2019-02-26
STITCH-1996: add reactive rebinding of services on user switch

adamchel:
This should cover 
* STITCH-1996 Add reactive user rebinding of services on switch
    * This also adds the new `StitchAuthDelegate` methods (these are the enhanced `StitchAuthListener` changes from the other SDKs) and integration tests for those methods
        * I noticed that the tests for this listener on Android were not very robust, so I opened the ticket STITCH-2570 to make those tests more robust.
* STITCH-2282 DataSynchronizer should make initialization async
    * I’m putting this up with the caveat that asynchronous data synchronizer initialization is not quite complete. Some calls that were once synchronous may now block, and there is the possibility of nil pointer accesses in edge cases we’ve missed.
    * These are not urgent concerns since the multi-user functionality works, but it’s something we should address soon, probably when Jason gets back so we can discuss the necessary API changes. I filed STITCH-2558 and STITCH-2559 to track this.
* STITCH-2264 Multiplex local client on app ID
    * Since each app ID and user ID has its own data directory, I also removed the instance key from the `sync_user_` and `sync_undo_` database names, so now user-defined db and coll names can be longer (the instance key is staying on the `sync_config_` db name since the namespace is not defined by user namespaces, but it’s not strictly necessary. I kept it because it may help us in the future with troubleshooting)

When this review is approved, I will manually squash my commits into one so that Jason’s foundational work on this PR is recognized in a commit in the git history of the SDK.


Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Internal/CoreRemoteMongoClient.swift
```diff
@@ -27,4 +33,36 @@ public class CoreRemoteMongoClient {
...
+    private func onAuthEvent(_ authEvent: AuthRebindEvent) {
+        switch authEvent {
+        case .userRemoved(let removedUser):
+            let userId = removedUser.id
+            let key = appInfo.clientAppID + "/\(userId)"
+            try? CoreLocalMongoDBService.shared.deleteDatabase(
+                withKey: key,
+                withDBPath: "\(appInfo.dataDirectory.path)/local_mongodb/0/\(key)")
+        case .activeUserChanged(let currentActiveUser, _):
+            if lastActiveUserId != appInfo.authMonitor.activeUserId {
```
tkaye407: Seems strange that you are not just getting the new authInfo from currentActiveUser, but I guess it shouldnt be a problem?

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Internal/CoreRemoteMongoClient.swift
```diff
@@ -27,4 +33,36 @@ public class CoreRemoteMongoClient {
...
+                self.lastActiveUserId = appInfo.authMonitor.activeUserId
+                if currentActiveUser != nil {
+                    self.dataSynchronizer.reinitialize(appInfo: appInfo)
+                } else {
+                    self.dataSynchronizer.stop()
+                }
```
tkaye407: When would currentActiveUser be null?

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/Internal/ThreadSafeMongoMobile.swift
```diff
@@ -45,110 +48,166 @@ class ThreadSafeMongoDatabase {
...
+        return try lock.write {
+            return try underlyingCollection().replaceOne(filter: filter, replacement: replacement, options: options)
+        }
     }
 
     @discardableResult
     func updateOne(filter: Document, update: Document, options: UpdateOptions? = nil) throws -> UpdateResult? {
+        return try lock.write {
         return try underlyingCollection().updateOne(filter: filter, update: update, options: options)
+        }
```
tkaye407: (nit) Indentation

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/Internal/ThreadSafeMongoMobile.swift
```diff
@@ -45,110 +48,166 @@ class ThreadSafeMongoDatabase {
...
+            return try underlyingCollection().find()
+        }
     }
 
     func find(_ filter: Document, options: FindOptions? = nil) throws -> MongoCursor<T> {
-        return try underlyingCollection().find(filter, options: options)
+        return try lock.write {
+            return try underlyingCollection().find(filter, options: options)
+        }
     }
```
tkaye407: Depending on which PR goes in first, I think my findOne() change will need to be editted

Core/StitchCoreAdminClient/Sources/StitchCoreAdminClient/StitchAdminAuth.swift
```diff
@@ -22,7 +22,7 @@ public final class StitchAdminAuth: CoreStitchAuth<StitchAdminUser> {
         return info
     }
 
-    public final override func onAuthEvent() {
-        // do nothing
+    public final override func dispatchAuthEvent(_ authEvent: AuthRebindEvent) {
+
     }
```
tkaye407: I personally like when these empty functions are only one line like 

 public final override func dispatchAuthEvent(_ authEvent: AuthRebindEvent) { } 

I feel like we use that pattern elsewhere (esp in JS sdk) but if that is not the pattern we want dont worry / let me know.

Core/StitchCoreAdminClient/Sources/StitchCoreAdminClient/StitchAdminUser.swift
```diff
@@ -28,12 +28,12 @@ public final class StitchAdminUser: CoreStitchUser {
...
      * A string describing the type of authentication provider used to log in as this user.
      */
-    public private(set) var loggedInProviderType: StitchProviderType
+    public private(set) var loggedInProviderType: StitchProviderType?
 
     /**
      * The name of the authentication provider used to log in as this user.
      */
-    public private(set) var loggedInProviderName: String
+    public private(set) var loggedInProviderName: String?
```
tkaye407: (nit) Curiosity, why did these have to become optional?

Core/StitchCoreSDK/Sources/StitchCoreSDK/Auth/Internal/CoreStitchAuth+AuthStorage.swift
```diff
@@ -82,25 +82,33 @@ extension CoreStitchAuth {
...
             if let newAuthInfo = activeUserAuthInfo {
                 try writeActiveUserAuthInfoToStorage(activeAuthInfo: newAuthInfo)
             }
+            dispatchAuthEvent(.activeUserChanged(currentActiveUser: nil,
+                                                 previousActiveUser: previousUser))
+
             return
         }
 
         // Otherwise make the user, if makeStitchUser() fails then something is very wrong
```
tkaye407: probably remove this comment

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Internal/CoreRemoteMongoClient.swift
adamchel: `currentActiveUser` is not available here, as this is a service client, not the `StitchAuth`. the `authMonitor` serves as the reference to auth information for this service client

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Internal/CoreRemoteMongoClient.swift
adamchel: If you log out, the active user changes from non-nil to nil. When you log out, we just want to stop the data synchronizer rather than reinitialize it

Core/StitchCoreAdminClient/Sources/StitchCoreAdminClient/StitchAdminAuth.swift
adamchel: I agree that it's a little cleaner. We did the other way because in Java the linter doesn't like it on a single line, but for Swift I'll try to start using the single-line approach.

Core/StitchCoreAdminClient/Sources/StitchCoreAdminClient/StitchAdminUser.swift
adamchel: Good q, I'm not sure. this was from Jason's commit. I'll investigate a little more and see if they need to be optional

Core/StitchCoreSDK/Sources/StitchCoreSDK/Auth/Internal/CoreStitchAuth+OtherAuthActions.swift
```diff
@@ -117,20 +114,19 @@ extension CoreStitchAuth {
...
     public func listUsersInternal() -> [TStitchUser] {
-        var list: [TStitchUser] = []
-        for userInfo in self.allUsersAuthInfo {
-            if let newUser = makeStitchUser(withAuthInfo: userInfo) {
-                list.append(newUser)
-            }
-        }
-
-        return list
+        return self.allUsersAuthInfo.compactMap({
```
tkaye407: neat

Darwin/Services/StitchRemoteMongoDBService/StitchRemoteMongoDBServiceTests/SyncIntTests.swift
```diff
@@ -2064,6 +2108,101 @@ class SyncIntTests: BaseStitchIntTestCocoaTouch {
...
+
+        // configure Sync
+        coll.configure(conflictHandler: failingConflictHandler)
+
+        let doc1Id = coll.insertOne(&docToInsertUser1)!.insertedId!
+        let doc1Filter = ["_id": doc1Id] as Document
+
+        // assert that the resolution is reflected locally, but not yet remotely
+        var expectedDocument = docToInsertUser1
+        XCTAssertEqual(expectedDocument.sorted(), coll.findOne(doc1Filter)!.sorted())
```
tkaye407: Isnt this call to findOne async and needs to be waited on?

adamchel: check out the `private extension RemoteMongoCollection` in this file. We added versions of all the CRUD methods that return synchronously

Darwin/StitchCore/StitchCore/Core/Auth/Internal/StitchAuthImpl+Delegation.swift
```diff
@@ -0,0 +1,91 @@
...
+        switch authEvent {
+        case .delegateRegistered:
+            delegate.onDelegateRegistered(auth: self)
+        case .userLoggedIn(let loggedInUser):
+            delegate.onUserLoggedIn(auth: self, loggedInUser: self.makeUser(loggedInUser))
+        case .userLoggedOut(let loggedOutUser):
+            delegate.onUserLoggedOut(auth: self, loggedOutUser: self.makeUser(loggedOutUser))
+        case .userLinked(let linkedUser):
+            delegate.onUserLinked(auth: self, linkedUser: self.makeUser(linkedUser))
+        case .activeUserChanged(let currentActiveUser, let previousActiveUser):
```
tkaye407: Why not pass the user i into the dispatch() function so that this is not done here

Darwin/StitchCore/StitchCore/Core/Auth/Internal/StitchAuthImpl.swift
```diff
@@ -303,44 +301,35 @@ internal final class StitchAuthImpl: CoreStitchAuth<StitchUserImpl>, StitchAuth
...
-     * Calls the `onAuthEvent` method of each registered `StitchAuthDelegate`.
+     * Dispatches the appropriate auth event method of each registered `StitchAuthDelegate`.
      *
      * - important: This is not meant to be invoked directly in this class. The `CoreStitchAuth` from which this
      *              class inherits will call this method when appropraite.
      */
-    public final override func onAuthEvent() {
-        self.delegates.enumerated().reversed().forEach { idx, delegateRef in
-            guard let delegate = delegateRef.value else {
+    public final override func dispatchAuthEvent(_ authEvent: AuthRebindEvent) {
```
tkaye407: Good note. I had to do the same thing below. Seems like Swift should fix this

Darwin/StitchCore/StitchCore/Core/Auth/Internal/StitchAuthImpl.swift
```diff
@@ -303,44 +301,35 @@ internal final class StitchAuthImpl: CoreStitchAuth<StitchUserImpl>, StitchAuth
...
+        // NOTE: this function is in this file rather than the +Delegation file because Swift does not support
+        //       overriding functions in extensions
```
tkaye407: How come we iterate through backwords?

Core/StitchCoreAdminClient/Sources/StitchCoreAdminClient/StitchAdminUser.swift
adamchel: I was able to make these non-optional by just adding nil guards to the `makeStitchUser` function. I'll push that change, but if there was a good reason to keep it optional happy to hear what @jsflax has to say when he gets back

Darwin/StitchCore/StitchCore/Core/Auth/StitchAuthDelegate.swift
```diff
@@ -13,7 +15,171 @@ public protocol StitchAuthDelegate: class {
...
+     * delegate.
+     *
+     * - parameters:
+     *    - auth: The instance of `StitchAuth` where the the delegate was registered.
+     *            It can be used to infer the current state of authentication.
+     */
+    func onDelegateRegistered(auth: StitchAuth)
+}
+
+// extensions that provide default implementations of the the auth delegate methods
```
tkaye407: Im confused by the pattern that is going on here. I understand delegates, but I don't get what is going on below with the extension if you have a minute to explain it to me.

Darwin/StitchCore/StitchCore/Core/Auth/Internal/StitchAuthImpl.swift
adamchel: this way, when we delete something while we're iterating through the list of delegates, we still see every item and don't accidentally miss one because the index wasn't updated when we deleted something

see: https://stackoverflow.com/questions/1582285/how-to-remove-elements-from-a-generic-list-while-iterating-over-it

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Internal/CoreRemoteMongoClient.swift
tkaye407: I meant the variable passed in to the method

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Internal/CoreRemoteMongoClient.swift
tkaye407: Sorry,, saw this later on and realized it was a dumb question

Darwin/StitchCore/StitchCore/Core/Auth/Internal/StitchAuthImpl.swift
tkaye407: interesting, thanks

Darwin/StitchCore/StitchCore/Core/Auth/StitchAuthDelegate.swift
adamchel: When people implement `StitchAuthDelegate`, we don't want them to be required to implement every single possible method (i.e., not all users will need to implement somtehing for `onUserAdded` or `onUserRemoved`, some users might care only about `onActiveUserChanged`).

By adding these extensions to the protocol, classes that implement the protocol will have these implementations by default.

It's why in the StitchAuthDelegateIntTests, we're able to do something like this:

```swift
        class TestAuthDelegate: StitchAuthDelegate {
            let work = DispatchGroup()
            var numCalls = 0

            func onUserAdded(auth: StitchAuth, addedUser: StitchUser) {
                XCTAssertNotNil(addedUser.id)
                numCalls += 1
                work.leave()
            }
        }
```

See https://docs.swift.org/swift-book/LanguageGuide/Protocols.html#ID521 for more context.

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Internal/CoreRemoteMongoClient.swift
adamchel: I didn't design this so I can't speak to it in depth, but from what I can tell, it's about keeping ` CoreRemoteMongoClient` simple, and not giving it too much state to maintain. 

The appInfo maintains a reference to the app client and auth but as an `AuthMonitor` rather than a full-fledged app client. This gives the CoreRemoteMongoClient only the state it needs to do its job, and nothing more.

@jsflax can probably speak more to this when he's back

Darwin/StitchCore/StitchCore/Core/Auth/Internal/StitchAuthImpl+Delegation.swift
adamchel: This is converting the `CoreStitchUser` from the `AuthRebindEvent` to a `StitchUser` for user consumption, so the user can't be passed in.

Darwin/Services/StitchRemoteMongoDBService/StitchRemoteMongoDBServiceTests/SyncIntTests.swift
tkaye407: Oh wow, i see. I just pushed a new change to my findOne() that changes the internal findOne() in SyncIntTests.

Darwin/StitchCore/StitchCore/Core/Auth/StitchAuthDelegate.swift
tkaye407: Yeah, I just throught that protocols are only method declarations

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Internal/CoreRemoteMongoClient.swift
jsflax: Correct.

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/Internal/ThreadSafeMongoMobile.swift
```diff
@@ -3,6 +3,9 @@ import StitchCoreSDK
 import StitchCoreLocalMongoDBService
 import Foundation
 
+// static global lock for all threads
+private let lock = ReadWriteLock(label: "local_mdb_lock")
```
jsflax: Nice.

PR #149, opened 2019-03-11
STITCH-2559

adamchel:
This makes all the previously blocking sync calls async, and I also drive-by'ed the $in optimizaiton (STITCH-2585).

I did not yet do the optimization that sets is stale for a whole namespace rather than per document, so this will still be relatively slow on disconnect/reconnect until that optimization is made.

I've also moved the CRUD operations from `Sync.swift` into a `Sync+CRUD.swift` to keep the files small.


Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/Internal/DataSynchronizer.swift
```diff
@@ -2279,10 +2279,10 @@ public class DataSynchronizer: NetworkStateDelegate, FatalErrorListener {
 
     private func latestStaleDocumentsFromRemote(nsConfig: NamespaceSynchronization,
                                                 staleIds: Set<AnyBSONValue>) throws -> [Document] {
-        let ids = staleIds.map { ["_id": $0.value ] as Document }
+        let ids = staleIds.map { $0.value }
         guard ids.count > 0 else { return [] }
```
jsflax: [nit] This guard can be one line earlier.

PR #155, opened 2019-03-27
STITCH-2653 Add instrumented network transport

adamchel:
This adds the instrumented Foundation transport for Swift.

To enable this, I had to slightly refactor the original Foundation HTTP transport, and I also added an internal property to the raw SSE stream returned from the stream request, which contains the initial response to the stream request so that it and its headers are counted along with the stream content.

I also removed the `StitchCoreTests.swift` file in `StitchCoreSDKTests` and moved its contents to the `StitchCoreTestUtils`, so that it can be shared by all test packages. This involved creating some new files, and making `Swifter` a dependency of multiple pods in our `Podfile`.


Darwin/Services/StitchRemoteMongoDBService/StitchRemoteMongoDBServiceTests/FoundationInstrumentedHTTPTransport.swift
```diff
@@ -0,0 +1,129 @@
+import Foundation
+import MongoSwift
+@testable import StitchCoreSDK
+
+/**
+ * A basic implementation of the `Transport` protocol using the `URLSession.dataTask` method in the Foundation library.
+ */
+public class FoundationInstrumentedHTTPTransport: FoundationHTTPTransport {
+    private var bytesUploadedForNormalRequests: Int64 = 0
+    private var bytesDownloadedForNormalRequests: Int64 = 0
```
jsflax: [nit] Can make both of these unsigned– use your best judgment for if that should be 32 or 64 bits.

Darwin/Services/StitchRemoteMongoDBService/StitchRemoteMongoDBServiceTests/FoundationInstrumentedHTTPTransportIntTests.swift
```diff
@@ -0,0 +1,275 @@
...
+    let streamEndpoint = "/sse"
+    let streamLines = [
+        "'And oh, what a terrible country it is!",
+        "Nothing but thick jungles infested by the most" +
+        "dangerous beasts in the entire world –",
+        "hornswogglers and snozzwangers and those terrible" +
+        "wicked whangdoodles.",
+        "A whangdoodle would eat ten Oompa-Loompas for " +
+        "breakfast and come galloping back for a second helping.'"
+    ]
```
jsflax: I'm glad this has stayed alive 😸

Darwin/Services/StitchRemoteMongoDBService/StitchRemoteMongoDBServiceTests/FoundationInstrumentedHTTPTransport.swift
adamchel: Interestingly enough, Foundation gives the byte counts as a signed int:
https://github.com/phracker/MacOSX-SDKs/blob/master/MacOSX10.9.sdk/System/Library/Frameworks/Foundation.framework/Versions/C/Headers/NSURLSession.h#L243

I'm going to leave this as is, because it feels unsafe to cast a signed long to an unsigned long from a library that explicitly gives a signed long

PR #158, opened 2019-04-01
STITCH-2654: Swift performance testing harness

tkaye407:
I left some of the logging in here in case I need to debug and because I assume I will have to make a few changes after you all look at it. For now, a few things to note

1. The networkMonitor and transport should be fully functional for both prod and local 
2. Prod tests still need to have the Stitch Server running in the background. Although it doesn't do anything with it, it seems to randomly fail with no explanation when the stitch server is not running. I was not sure how important this is though. 
3. We now have the ThrowingCallBackJoiner which on call to value() will throw if there was an error. This allows us to catch errors in the harness
4. Errors now update the test document to have a status of "failed" and adds a description of the failure. 
5. I haven't seen them in a while, but there appears to be some flakey-ness in the setup/teardown logic of local tests where it occasionally fails. I haven't seen it in the last 20 or so runs, but I would appreciate it if you take special note of the setup/teardown logic as you go through this PR.


Darwin/Services/StitchRemoteMongoDBService/StitchRemoteMongoDBServiceTests/SyncPerformanceContext.swift
```diff
@@ -0,0 +1,307 @@
...
+import StitchCore
+import StitchCoreSDK
+import StitchCoreAdminClient
+import StitchDarwinCoreTestUtils
+@testable import StitchCoreRemoteMongoDBService
+import StitchCoreLocalMongoDBService
+@testable import StitchRemoteMongoDBService
+
+class SyncPerformanceContext {
+    var client: StitchAppClient?
```
jsflax: [nit] all of these vars should have private(set)

Darwin/Services/StitchRemoteMongoDBService/StitchRemoteMongoDBServiceTests/SyncPerformanceContext.swift
```diff
@@ -0,0 +1,307 @@
...
+
+        coll?.sync.proxy.dataSynchronizer.isSyncThreadEnabled = false
+        coll?.sync.proxy.dataSynchronizer.stop() // Failing here occaisonally
+        print("PerfLog: Context: Done with initialization")
+    }
+
+    func runSingleIteration(numDocs: Int,
+                            docSize: Int,
+                            testDefinition: TestDefinition,
+                            customSetup: TestDefinition? = nil,
```
jsflax: [nit] give different typealiases to these variables. They should also be named differently (`custom` is redundant), and instead of dealing with nullability, just make them empty closures.

Darwin/Services/StitchRemoteMongoDBService/StitchRemoteMongoDBServiceTests/SyncPerformanceContext.swift
```diff
@@ -0,0 +1,307 @@
...
+    }
+
+    func resume() {
+        threadData = []
+        cpuData = []
+        memoryData = []
+        timeSource.resume()
+    }
+    // swiftlint:disable large_tuple
+    func suspend() -> (Double, Double, Double) {
```
jsflax: [nit] Use a typealias or struct

Darwin/Services/StitchRemoteMongoDBService/StitchRemoteMongoDBServiceTests/SyncPerformanceContext.swift
```diff
@@ -0,0 +1,307 @@
...
+internal class MetricsCollector {
+    let timeSource: DispatchSourceTimer
+    private var threadData: [Double] = []
+    private var cpuData: [Double] = []
+    private var memoryData: [Double] = []
+
+    init(timeInterval: TimeInterval) {
+        timeSource = DispatchSource.makeTimerSource()
+        timeSource.schedule(deadline: .now(), repeating: timeInterval)
+        timeSource.setEventHandler(handler: { [weak self] in
```
jsflax: [nit] `guard` against `self` here

Darwin/Services/StitchRemoteMongoDBService/StitchRemoteMongoDBServiceTests/SyncPerformanceIntTestHarness.swift
```diff
@@ -0,0 +1,341 @@
...
+
+        if !(outputClient?.auth.isLoggedIn ?? false) {
+            let apiKeyOpt = ProcessInfo.processInfo.environment["PERF_IOS_API_KEY"]
+            guard let apiKey = apiKeyOpt else {
+                XCTFail("No proper iOS-API-Key for stitch perf project gixen")
+                return
+            }
+
+            outputClient?.auth.login(withCredential: UserAPIKeyCredential(withKey: apiKey), callbackJoiner.capture())
+
```
jsflax: [nit] extra newline

Darwin/Services/StitchRemoteMongoDBService/StitchRemoteMongoDBServiceTests/SyncPerformanceIntTestHarness.swift
```diff
@@ -0,0 +1,341 @@
...
+        outputMongoClient = try! outputClient?.serviceClient(fromFactory: remoteMongoClientFactory,
+                                                             withName: "mongodb-atlas")
+        outputColl = outputMongoClient?.db(stitchOutputDbName).collection(stitchOutputCollName)
+    }
+
+    override func tearDown() {
+        super.tearDown()
+    }
+
+    func setUpIter() {
```
jsflax: [q][nit] what are these empty methods?

Darwin/Services/StitchRemoteMongoDBService/StitchRemoteMongoDBServiceTests/SyncPerformanceIntTestHarness.swift
```diff
@@ -0,0 +1,341 @@
...
+            Failed on iteration \(iteration) of \(testParams.numIters)
+            with error \(String(describing: error))
+            """
+
+            print("PerfLog: Harness Error: \(failureMessage)")
+            if testParams.outputToStitch {
+                _ = outputColl?.updateOne(filter: ["_id": resultId],
+                                          update: ["$set": ["status": "Failed",
+                                                            "failureReason": failureMessage] as Document])
+            XCTFail(failureMessage)
```
jsflax: [style] indentation

Darwin/Services/StitchRemoteMongoDBService/StitchRemoteMongoDBServiceTests/SyncPerformanceIntTestHarness.swift
```diff
@@ -0,0 +1,341 @@
...
+        self.docSizes = docSizes
+        self.dataProbeGranularityMs = dataProbeGranularityMs
+        self.numOutliersEachSide = numOutliersEachSide
+        self.stitchHostName = stitchHostName
+        self.outputToStdOut = outputToStdOut
+        self.outputToStitch = outputToStitch
+        self.preserveRawOutput = preserveRawOutput
+    }
+
+    func toBson() -> Document {
```
jsflax: [nit] this should be a lazy var

Darwin/Services/StitchRemoteMongoDBService/StitchRemoteMongoDBServiceTests/SyncPerformanceIntTestHarness.swift
```diff
@@ -0,0 +1,341 @@
...
+            "numOutliersEachSide": numOutliersEachSide,
+            "stitchHostName": stitchHostName, // Todo: fix this
+            "date": Date(),
+            "sdk": "ios",
+            "results": []
+        ]
+    }
+}
+
+struct DoubleDataBlock {
```
jsflax: [nit] Calling it a `Double` data block is redundant. Just call it a data block.

Darwin/Services/StitchRemoteMongoDBService/StitchRemoteMongoDBServiceTests/SyncPerformanceIntTestHarness.swift
```diff
@@ -0,0 +1,341 @@
...
+    var mean: Double = 0.0
+    var median: Double = 0.0
+    var max: Double = 0.0
+    var min: Double = 0.0
+    var stdDev: Double = 0.0
```
jsflax: [nit] all of these should be `let`s

Darwin/Services/StitchRemoteMongoDBService/StitchRemoteMongoDBServiceTests/SyncPerformanceIntTestHarness.swift
```diff
@@ -0,0 +1,341 @@
...
+                median = newData[middle]
+            }
+
+            mean = newData.reduce(0.0, +) / Double(newData.count)
+            let sumOfSquared = newData.map {($0 - mean) * ($0 - mean)}.reduce(0, +)
+            stdDev = sqrt(sumOfSquared / Double(newData.count))
+        }
+    }
+
+    func toBson() -> Document {
```
jsflax: [nit] this should be a lazy var

Darwin/Services/StitchRemoteMongoDBService/StitchRemoteMongoDBServiceTests/SyncPerformanceIntTestHarness.swift
```diff
@@ -0,0 +1,341 @@
...
+internal struct IterationResult {
+    let time: Double
+    let networkSentBytes: Double
+    let networkReceivedBytes: Double
+    let cpu: Double
+    let memory: Double
+    let disk: Double
+    let threads: Double
+
+    init(time: Double,
```
jsflax: [nit] you can use the default constructor instead of this

Darwin/Services/StitchRemoteMongoDBService/StitchRemoteMongoDBServiceTests/SyncPerformanceIntTests.swift
```diff
@@ -0,0 +1,113 @@
+//
+//  SyncPerformanceIntTests.swift
+//  StitchRemoteMongoDBServiceTests
+//
+//  Created by Tyler Kaye on 3/29/19.
+//  Copyright © 2019 MongoDB. All rights reserved.
+//
```
jsflax: [style] remove template comment

Darwin/Services/StitchRemoteMongoDBService/StitchRemoteMongoDBServiceTests/SyncPerformanceIntTests.swift
```diff
@@ -0,0 +1,113 @@
...
+class SyncPerformanceIntTests: BaseStitchIntTestCocoaTouch {
+    let harness = SyncPerformanceIntTestHarness()
+    static let runId = ObjectId()
+    let joiner = ThrowingCallbackJoiner()
+
+    override func setUp() {
+        super.setUp()
+    }
+
+    override func tearDown() {
```
jsflax: [nit] remove unnecessary override

Darwin/Services/StitchRemoteMongoDBService/StitchRemoteMongoDBServiceTests/SyncPerformanceContext.swift
```diff
@@ -0,0 +1,307 @@
...
+            let svc = try! harness.addService(toApp: app.1, withType: "mongodb", withName: "mongodb1",
+                                         withConfig: ServiceConfigs.mongodb(name: "mongodb1", uri: harness.mongodbUri))
+            let rule = RuleCreator.mongoDb(database: dbName, collection: collName,
+                                           roles: [RuleCreator.Role(read: true, write: true)],
+                                           schema: RuleCreator.Schema())
+            _ = try! harness.addRule(toService: svc.1, withConfig: rule)
+
+            client = try! harness.appClient(forApp: app.0, withTransport: transport)
+            client?.auth.login(withCredential: AnonymousCredential(), joiner.capture())
+            let _: Any? = try joiner.value()
```
jsflax: [style] unnecessary type declaration

Darwin/Services/StitchRemoteMongoDBService/StitchRemoteMongoDBServiceTests/SyncPerformanceContext.swift
```diff
@@ -0,0 +1,307 @@
...
+
+    var dbName = ""
+    var collName = ""
+    var userId = ""
+
+    let joiner = ThrowingCallbackJoiner()
+    let streamJoiner = StreamJoiner()
+    let testParams: TestParams?
+    let transport: FoundationInstrumentedHTTPTransport?
+    let harness: SyncPerformanceIntTestHarness?
```
jsflax: [minor] none of the above lets should be optional. If you feel they must, then this was likely designed wrong

Darwin/Services/StitchRemoteMongoDBService/StitchRemoteMongoDBServiceTests/SyncPerformanceIntTestHarness.swift
```diff
@@ -0,0 +1,341 @@
...
+            outputClient = try Stitch.appClient(forAppID: stitchOutputAppName)
+        } catch {
+            let config = StitchAppClientConfigurationBuilder()
+                .with(transport: networkTransport)
+                .with(networkMonitor: networkMonitor).build()
+            outputClient = try! Stitch.initializeAppClient(withClientAppID: stitchOutputAppName, withConfig: config)
+        }
+
+        if !(outputClient?.auth.isLoggedIn ?? false) {
+            let apiKeyOpt = ProcessInfo.processInfo.environment["PERF_IOS_API_KEY"]
```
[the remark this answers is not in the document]
tkaye407: No, I intentionally made this all caps. Seems to be the thing to do in the rest of the code base no?

Darwin/Services/StitchRemoteMongoDBService/StitchRemoteMongoDBServiceTests/SyncPerformanceContext.swift
tkaye407: Kind of confused. Why give them different typealiases?  Also if I make them empty closures then they dont conform to the type alias which swift complains about

jsflax: Because even though the type definitions happen to be the same, they have semantically different uses.

To compile default args, you likely need to add empty params. So not:
`{}`
but:
`{ _, _, _ in }`
or something

Darwin/Services/StitchRemoteMongoDBService/StitchRemoteMongoDBServiceTests/SyncPerformanceIntTestHarness.swift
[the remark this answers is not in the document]
jsflax: Ah yes. We'll need to add an objective-c header file as well.

Darwin/Services/StitchRemoteMongoDBService/StitchRemoteMongoDBServiceTests/SyncPerformanceContext.swift
tkaye407: If you dont have :Any? then .value() fails because it cannot infer the type

Darwin/Services/StitchRemoteMongoDBService/StitchRemoteMongoDBServiceTests/SyncPerformanceIntTestHarness.swift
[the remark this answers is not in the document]
tkaye407: I tried this and it wasnt working, Ill try again later

Darwin/Services/StitchRemoteMongoDBService/StitchRemoteMongoDBServiceTests/SyncPerformanceIntTestHarness.swift
tkaye407: I cant do this in case you set numOutliers to be more than the number of iterations I just return these values

Darwin/Services/StitchRemoteMongoDBService/StitchRemoteMongoDBServiceTests/ProductionPerformanceContext.swift
```diff
@@ -0,0 +1,59 @@
...
+
+        dbName = harness.stitchTestDbName
+        collName = harness.stitchTestCollName
+        client = harness.outputClient
+        mongoClient = harness.outputMongoClient
+        coll = mongoClient.db(dbName).collection(collName)
+        userId = client.auth.currentUser?.id ?? "No User"
+
+        coll.sync.proxy.dataSynchronizer.isSyncThreadEnabled = false
+        coll.sync.proxy.dataSynchronizer.stop() // Failing here occaisonally
```
jsflax: [q] Is this still failing here occasionally? If so, we should fix it.

Darwin/Services/StitchRemoteMongoDBService/StitchRemoteMongoDBServiceTests/SyncPerformanceIntTests.swift
```diff
@@ -0,0 +1,83 @@
...
+            print("PerfLog: Test: \(numDoc) docs of size \(docSize)")
+            let docs = getDocuments(numDocs: numDoc, docSize: docSize)
+            ctx.coll.insertMany(docs, joiner.capture())
+            let _: Any? = try joiner.value()
+        })
+    }
+
+    func getDocuments(numDocs: Int, docSize: Int) -> [Document] {
+        var docs: [Document] = []
+        for iter in 1...numDocs {
```
jsflax: [nit] You're better off mapping the range into the document array, and having this method throw (or fatalError). If we can't create documents we can't really test performance. Eating that error is useless

Darwin/Services/StitchRemoteMongoDBService/StitchRemoteMongoDBServiceTests/ProductionPerformanceContext.swift
tkaye407: Ill remove this comment, I havent seen it fail in the past few days

tkaye407: I must have fixed it at some point while fixing something else :)

Darwin/Services/StitchRemoteMongoDBService/StitchRemoteMongoDBServiceTests/LocalPerformanceContext.swift
```diff
@@ -0,0 +1,71 @@
...
+import MongoSwift
+import StitchCore
+import StitchCoreSDK
+import StitchCoreAdminClient
+import StitchDarwinCoreTestUtils
+@testable import StitchCoreRemoteMongoDBService
+import StitchCoreLocalMongoDBService
+@testable import StitchRemoteMongoDBService
+
+class LocalPerformanceContext: SyncPerformanceContext {
```
adamchel: [naming] I think this should `LocalPerformanceTestContext`

Darwin/Services/StitchRemoteMongoDBService/StitchRemoteMongoDBServiceTests/LocalPerformanceContext.swift
```diff
@@ -0,0 +1,71 @@
+// swiftlint:disable force_try
```
adamchel: [aside] I know we do this everywhere in our tests, so I think it's fine for now but I thinkgoing forward we should try to avoid this. Right now, a lot of our tests fail with no explanation because of force-try failures, and more carefully catching errors in test setup would help us resolve strange issues.

Darwin/Services/StitchRemoteMongoDBService/StitchRemoteMongoDBServiceTests/SyncIntTests.swift
```diff
@@ -15,7 +15,7 @@ import StitchCoreLocalMongoDBService
 
 private let waitTimeout = UInt64(1e+10)
 
-private extension Document {
+internal extension Document {
```
adamchel: [minor] Since these extensions and classes are no longer unique to this file, we should throw them them into separate files. It'll be easier to keep track of them.

The convention for naming extension files is to do stuff like `Document+Sorted.swift`, `RemoteMongoCollection+SynchronousCRUD.swift`.

StreamJoiner and SyncTestContext can be named normally.

Darwin/Services/StitchRemoteMongoDBService/StitchRemoteMongoDBServiceTests/SyncPerformanceContext.swift
```diff
@@ -0,0 +1,232 @@
+import MongoSwift
+import StitchCore
+import StitchCoreSDK
+import StitchCoreAdminClient
+import StitchDarwinCoreTestUtils
+@testable import StitchCoreRemoteMongoDBService
+import StitchCoreLocalMongoDBService
+@testable import StitchRemoteMongoDBService
+
+protocol SyncPerformanceContext {
```
adamchel: [naming] `SyncPerformanceTestContext`

Darwin/Services/StitchRemoteMongoDBService/StitchRemoteMongoDBServiceTests/SyncPerformanceContext.swift
```diff
@@ -0,0 +1,232 @@
...
+         transport: FoundationInstrumentedHTTPTransport) throws
+
+    func tearDown() throws
+}
+
+extension SyncPerformanceContext {
+    func runSingleIteration(numDocs: Int,
+                            docSize: Int,
+                            testDefinition: TestDefinition,
+                            setup: SetupDefinition,
```
adamchel: [naming] let's call these `customSetup` and `customTeardown` so it's more clear what they do

Darwin/Services/StitchRemoteMongoDBService/StitchRemoteMongoDBServiceTests/SyncPerformanceContext.swift
```diff
@@ -0,0 +1,232 @@
...
+        timeSource.cancel()
+        return PointInTimeMetrics(threadData: threadData.reduce(0.0, +) / Double(threadData.count),
+                                  cpuData: cpuData.reduce(0.0, +) / Double(cpuData.count),
+                                  memoryData: memoryData.reduce(0.0, +) / Double(memoryData.count))
+    }
+}
+
+// Helper functions to get usage stats
+// Solution taken from: https://stackoverflow.com/questions/8223348/ios-get-cpu-usage-from-application
+private func cpuUsage() -> (Int, Double) {
```
adamchel: looks like these comments were not actually resolved

Darwin/Services/StitchRemoteMongoDBService/StitchRemoteMongoDBServiceTests/SyncPerformanceContext.swift
```diff
@@ -0,0 +1,307 @@
...
+
+    if kerr == KERN_SUCCESS {
+        return Double(taskInfo.resident_size)
+    } else {
+        print("Error getting memory")
+        return 0.0
+    }
+}
+
+internal class MetricsCollector {
```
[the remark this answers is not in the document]
adamchel: looks like this wasn't resolved

Darwin/Services/StitchRemoteMongoDBService/StitchRemoteMongoDBServiceTests/SyncPerformanceContext.swift
```diff
@@ -0,0 +1,232 @@
...
+    }
+
+    func waitForAllStreamsOpen() {
+        while !coll.sync.proxy.dataSynchronizer.allStreamsAreOpen {
+            print("waiting for all streams to open before doing sync pass")
+            sleep(1)
+        }
+    }
+
+    func streamAndSync() throws {
```
adamchel: [note] In the actual performance tests, we won't want to call this, because it wouldn't be an accurate simulation of real behavior. It's fine for your placeholder tests though.

Darwin/Services/StitchRemoteMongoDBService/StitchRemoteMongoDBServiceTests/SyncPerformanceIntTests.swift
```diff
@@ -0,0 +1,95 @@
...
+@testable import StitchRemoteMongoDBService
+
+import Foundation
+
+class SyncPerformanceIntTests: BaseStitchIntTestCocoaTouch {
+    let harness = SyncPerformanceIntTestHarness()
+    static let runId = ObjectId()
+    let joiner = ThrowingCallbackJoiner()
+
+    /* func testFailure() {
```
adamchel: [q] why is this commented out?

Darwin/Services/StitchRemoteMongoDBService/StitchRemoteMongoDBServiceTests/SyncPerformanceIntTests.swift
```diff
@@ -0,0 +1,95 @@
...
+import StitchCoreSDK
+import StitchCoreAdminClient
+import StitchDarwinCoreTestUtils
+@testable import StitchCoreRemoteMongoDBService
+import StitchCoreLocalMongoDBService
+@testable import StitchRemoteMongoDBService
+
+import Foundation
+
+class SyncPerformanceIntTests: BaseStitchIntTestCocoaTouch {
```
adamchel: [q] I thought we decided we didn't want to inherit from BaseStitchIntTestCocoaTouch. What changed? This is part of why you're getting errors when the local server is not running for prod tests.

Darwin/Services/StitchRemoteMongoDBService/StitchRemoteMongoDBServiceTests/SyncPerformanceIntTestHarness.swift
```diff
@@ -0,0 +1,363 @@
...
+    }
+
+    override func tearDown() {
+        super.tearDown()
+    }
+
+    // swiftlint:disable function_body_length
+    func runPerformanceTestWithParameters(testParams: TestParams,
+                                          testDefinition: TestDefinition,
+                                          customSetup: SetupDefinition = { _, _, _ in },
```
adamchel: [naming] let's call these `beforeEach` and `afterEach` to match Java

Darwin/Services/StitchRemoteMongoDBService/StitchRemoteMongoDBServiceTests/SyncPerformanceIntTestHarness.swift
```diff
@@ -0,0 +1,363 @@
...
+typealias TestDefinition = (_ ctx: SyncPerformanceContext, _ numDoc: Int, _ docSize: Int) throws -> Void
+typealias SetupDefinition = (_ ctx: SyncPerformanceContext, _ numDoc: Int, _ docSize: Int) throws -> Void
+typealias TeardownDefinition = (_ ctx: SyncPerformanceContext, _ numDoc: Int, _ docSize: Int) throws -> Void
+
+// This is how we want to do things once we create the new test scheme and import the .h file as an
+// objective-c bridging header
+// let testStitchAPIKey = PERF_IOS_API_KEY.isEmpty ?
+//    ProcessInfo.processInfo.environment["PERF_IOS_API_KEY"] : PERF_IOS_API_KEY
+
+class SyncPerformanceIntTestHarness: BaseStitchIntTestCocoaTouch {
```
adamchel: [q] is it still necessary to inherit this class?

Darwin/Services/StitchRemoteMongoDBService/StitchRemoteMongoDBServiceTests/SyncPerformanceIntTestHarness.swift
```diff
@@ -0,0 +1,363 @@
...
+        "docSize": docSize,
+        "status": "success",
+        "timeMs": time.asBson,
+        "networkSentBytes": networkSentBytes.asBson,
+        "networkReceivedBytes": networkReceivedBytes.asBson,
+        "cpu": cpu.asBson,
+        "memoryBytes": memory.asBson,
+        "diskBytes": disk.asBson,
+        "diskEfficiencyRatio": diskEfficiencyRatio.asBson,
+        "threads": threads.asBson
```
adamchel: [naming] seems like Jason (or someone) changed this on Android to "activeThreadCounts". We should make sure the Android and iOS param names match so there's a common data model between the two

Darwin/Services/StitchRemoteMongoDBService/StitchRemoteMongoDBServiceTests/SyncPerformanceContext.swift
jsflax: I disagree with that. `custom` is implied.

Darwin/Services/StitchRemoteMongoDBService/StitchRemoteMongoDBServiceTests/SyncPerformanceIntTestHarness.swift
jsflax: T'was me. `activeThreadCounts` is a bit more descriptive than "threads" 😄

Darwin/Services/StitchRemoteMongoDBService/StitchRemoteMongoDBServiceTests/SyncPerformanceContext.swift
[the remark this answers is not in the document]
tkaye407: We spoke about this in the channel. It is necessary to make this a class for right now

Darwin/Services/StitchRemoteMongoDBService/StitchRemoteMongoDBServiceTests/LocalPerformanceContext.swift
tkaye407: I agree, fixed this

Darwin/Services/StitchRemoteMongoDBService/StitchRemoteMongoDBServiceTests/SyncPerformanceContext.swift
tkaye407: lol, jason had me take the word custom out....

Darwin/Services/StitchRemoteMongoDBService/StitchRemoteMongoDBServiceTests/SyncPerformanceIntTests.swift
tkaye407: Because this is used to test failure and I didn't want to delete it but it would theoretically break evergreen if evergreen was running these tests properly

adamchel: Alright so add a comment explaining that

Darwin/Services/StitchRemoteMongoDBService/StitchRemoteMongoDBServiceTests/SyncPerformanceContext.swift
adamchel: alright fair enough. I still think we should name them `beforeEach` and `afterEach` in the other class though

Darwin/Services/StitchRemoteMongoDBService/StitchRemoteMongoDBServiceTests/SyncPerformanceIntTestHarness.swift
tkaye407: Here it sadly is

tkaye407:
Made a bunch of changes and tested this out a lot. The only problem I have found is that when running tests against prod the instrumented network transport only works for the first test run, which is a problem that ill continue to look into

Darwin/Services/StitchRemoteMongoDBService/StitchRemoteMongoDBServiceTests/SyncPerformanceIntTestHarness.swift
```diff
@@ -0,0 +1,364 @@
...
+        self.disk = disk
+        self.threads = threads
+    }
+}
+
+private class RunResults {
+    let numDocs: Int
+    let docSize: Int
+    let numOutliers: Int
+    let time: DataBlock
```
jsflax: [nit] should make the name more descriptive

Darwin/Services/StitchRemoteMongoDBService/StitchRemoteMongoDBServiceTests/SyncPerformanceIntTestHarness.swift
```diff
@@ -0,0 +1,364 @@
...
+    let networkSentBytes: DataBlock
+    let networkReceivedBytes: DataBlock
+    let cpu: DataBlock
+    let memory: DataBlock
+    let disk: DataBlock
```
jsflax: [nit] disk => diskUsage

Darwin/Services/StitchRemoteMongoDBService/StitchRemoteMongoDBServiceTests/SyncPerformanceIntTestHarness.swift
```diff
@@ -0,0 +1,364 @@
...
+    let diskEfficiencyRatio: DataBlock
+    let threads: DataBlock
```
jsflax: [nit] threads => activeThreadCount

Darwin/Services/StitchRemoteMongoDBService/StitchRemoteMongoDBServiceTests/SyncPerformanceTestContext.swift
```diff
@@ -0,0 +1,247 @@
...
+        $0.withMemoryRebound(to: integer_t.self, capacity: 1) {
+            task_info(mach_task_self_, task_flavor_t(MACH_TASK_BASIC_INFO), $0, &count)
+        }
+    }
+
+    if kerr == KERN_SUCCESS {
+        return Double(taskInfo.resident_size)
+    } else {
+        print("Error getting memory")
+        return 0.0
```
jsflax: [q][minor] Maybe this should be a fatal error? This would completely skew results if some tests have memory usage and others do not.

PR #161, opened 2019-04-08
STITCH-2655 Swift SDK: Automate the test harness into an xcode scheme

jsflax:
Includes updates to evergreen to run each test as a `task` within a `task_group`.
Notes:
- There seems to be a flaky test in `StitchCore`. I'm still investigating, and will include it in the scope of this ticket if possible, but it does not affect the review
- I did not divide the core tests into individual tasks because in their entirety they only take 60 seconds to run


Darwin/StitchCore/StitchCoreTests/Core/Auth/StitchAuthDelegateIntegrationTests.swift
```diff
@@ -33,7 +33,7 @@ class StitchAuthDelegateIntTests: BaseStitchIntTestCocoaTouch {
     }
 
     func waitForDelegate(work: DispatchGroup) {
-        guard work.wait(timeout: .now() + 10) == .success else {
+        guard work.wait(timeout: .now() + 20) == .success else {
```
dkaminsky: timeout is an absolute time?

Darwin/Services/StitchRemoteMongoDBService/StitchSyncPerformanceTests/StitchSyncPerformanceTests.swift
```diff
@@ -0,0 +1,33 @@
+//
+//  StitchSyncPerformanceTests.swift
+//  StitchSyncPerformanceTests
+//
+//  Created by Jason Flax on 06/04/2019.
+//  Copyright © 2019 MongoDB. All rights reserved.
```
adamchel: [nit] remove this copyright notice

Core/Services/StitchCoreRemoteMongoDBService/Tests/StitchCoreRemoteMongoDBServiceTests/CoreRemoteMongoCollectionUnitTests.swift
```diff
@@ -1142,15 +1142,15 @@ final class CoreRemoteMongoCollectionUnitTests: XCMongoMobileTestCase {
         exp.isInverted = true
         try expectStreamToClose(expectation: exp)
         coreRemoteMongoClient.onRebindEvent(AuthRebindEvent.userLoggedIn(loggedInUser: testUser))
-        wait(for: [exp], timeout: 0.1)
+        wait(for: [exp], timeout: 10)
```
adamchel: [q] the expectation is inverted, so shouldn't these be testing that the expectation never gets fulfilled? was this actually failing before? why did this need to change?

etc/read_ios_runtime.py
```diff
@@ -0,0 +1,8 @@
+
```
adamchel: [nit] remove this newline

Darwin/StitchCore/StitchCoreTests/Core/Auth/StitchAuthDelegateIntegrationTests.swift
adamchel: [q] Am I reading the docs right in that this is nanosecond precision? if it is, we should probably wait at least a second.

jsflax:
I haven't made performance a second task yet because there are still decisions that need to be made around that for both iOS and Android.

Core/Services/StitchCoreRemoteMongoDBService/Tests/StitchCoreRemoteMongoDBServiceTests/CoreRemoteMongoCollectionUnitTests.swift
jsflax: This test was flaky before. I assumed it was related to timeout.

Darwin/StitchCore/StitchCoreTests/Core/Auth/StitchAuthDelegateIntegrationTests.swift
jsflax: updated

PR #172, opened 2019-04-22
STITCH-2708 Swift SDK: Adopt Swift 5.0-compatible Swift Driver when it is available

jsflax:


jsflax:
TODO: Replace the callback joiner with the new await syntax I've added. I'll make it ticket.

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/Internal/DocumentVersionInfo.swift
```diff
@@ -64,7 +64,7 @@ final class DocumentVersionInfo {
         /// the GUID instance id of this version
         let instanceId: String
         /// the version counter of this version
-        let versionCounter: Int64
+        let versionCounter: Int
     }
```
tkaye407: Can you explain this? Why do we no longer have to specify it as a long?

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/Internal/DocumentVersionInfo.swift
```diff
@@ -64,7 +64,7 @@ final class DocumentVersionInfo {
         /// the GUID instance id of this version
         let instanceId: String
         /// the version counter of this version
-        let versionCounter: Int64
+        let versionCounter: Int
```
adamchel: [minor] I think this needs to be an Int64. (probably super unlikely, but If this ever got run on a 32-bit system, this would not behave correctly in the case that a document is updated more than 2 billion times)

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/Internal/DocumentVersionInfo.swift
jsflax: It makes it easier to work with when testing if it's an `Int`. Plus:
https://stackoverflow.com/questions/27440100/what-is-the-difference-between-int-and-int32-in-swift

Darwin/Examples/ToDoSync/ToDoSync/TodoTableViewController.swift
```diff
@@ -159,63 +154,67 @@ class TodoTableViewController:
...
+                                                    }
+                                            }
+                                        }
         }
     }
 
-    private func loggedIn() {
-        if listsCollection.sync.syncedIds.isEmpty {
-            listsCollection.sync.insertOne(document: TodoList(id: userId!)) { _ in }
+    private func loggedIn() { async {
```
adamchel: [broader discussion, does not need be addressed right now] if this hacky workaround stuff has to be done for usage of our API to be clean, we should perhaps rethink the API before GA. Maybe it makes sense to offer blocking versions of certain things 🤷‍♂️

Darwin/Services/StitchRemoteMongoDBService/StitchRemoteMongoDBServiceTests/RemoteMongoClientIntTests.swift
```diff
@@ -1716,11 +1725,25 @@ class RemoteMongoClientIntTests: BaseStitchIntTestCocoaTouch {
         )
 
         // should receive events for multiple documents being watched
+//        let doc2Oid = ObjectId()
```
adamchel: [nit] remove comment

Darwin/Services/StitchRemoteMongoDBService/StitchRemoteMongoDBServiceTests/RemoteMongoClientIntTests.swift
```diff
@@ -1716,11 +1725,25 @@ class RemoteMongoClientIntTests: BaseStitchIntTestCocoaTouch {
...
         let doc2: Document = [
             "_id": 42,
             "hello": "i am a number doc"
         ]
 
+        let aaaa = ["_id": 42] as Document
```
adamchel: [nit] remove debugging prints here

Darwin/Services/StitchRemoteMongoDBService/StitchRemoteMongoDBServiceTests/SyncIntTestUtilities.swift
```diff
@@ -27,8 +27,8 @@ internal class SyncIntTestUtilities {
         return document["__stitch_sync_version"] as! Document
     }
 
-    static internal func versionCounterOf(_ document: Document) -> Int64 {
-        return versionOf(document)["v"] as! Int64
+    static internal func versionCounterOf(_ document: Document) -> Int {
```
adamchel: [minor] should be Int64

Podfile
```diff
@@ -10,7 +10,11 @@ target :MockUtils do
 end
 
 def shared_pods
-    pod 'MongoSwift', '= 0.0.9'
+    pod 'MongoSwift', '= 0.1.0'
+end
+
+def mongo_mobile
+    pod 'MongoMobile', :git => 'https://github.com/mongodb/swift-mongo-mobile/', :branch => 'master'
```
adamchel: [minor] we should not point to master since this could break at any time without warning

Podfile
```diff
@@ -208,13 +212,13 @@ end
...
     project 'Darwin/Examples/ToDoSync/ToDoSync.xcodeproj'
-    pod 'MongoMobile', '= 0.0.7'
+    mongo_mobile
     pod 'Toast-Swift', '= 4.0.0'
     pod 'BEMCheckBox', '= 1.4.1'
 end
 
 target :StitchSDK do
-    pod 'MongoSwift', '= 0.0.9'
+    pod 'MongoSwift', '= 0.1.0'
```
adamchel: [nit] why not `shared_pods`?

Darwin/Services/StitchRemoteMongoDBService/StitchRemoteMongoDBServiceTests/SyncIntTestUtilities.swift
jsflax: See my response to Tyler's comment.

Darwin/Examples/ToDoSync/ToDoSync/TodoTableViewController.swift
jsflax: We should probably adopt either PromiseKit or Async/Await.

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/Internal/DocumentVersionInfo.swift
dkaminsky: Shouldn't this be consistent between implementations? -- in Java it's a long which is 64 bits on all architectures

jsflax: I've changed it back

dkaminsky: I mean, ostensibly it shouldn't matter since a swift vs java client would have different instance IDs and the likelihood of us maxing that out are nil... but still

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/Internal/DocumentVersionInfo.swift
dkaminsky: ditto

dkaminsky:
lgtm, seems Adam covered the rest - definitely get the ball rolling on evergreen fixes ASAP

PR #176, opened 2019-04-22
STITCH-2832 Swift SDK: Create Sync Frequency Associated Enum

jsflax:


Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Sync/Internal/SyncFrequency.swift
```diff
@@ -0,0 +1,28 @@
+import Foundation
+
+/**
+ * The frequency a user may choose for sync.
+ *
+ * Configurations should be done using the static initializers for the frequency
+ * type of choice, e.g.:
+ * sync.configure(frequency: .scheduled(timeInterval: .hours(1), isConnected: true))
```
adamchel: @tkaye407 or whoever ends up doing the builder pattern for this, let's make sure this snippet gets updated when we make that change

PR #189, opened 2019-07-04
STITCH-2989 Support additional watch() options in Swift SDK

adamchel:
Adds full-collection and match filter watch. Does not support compact events on the new watch options.

There's a drive-by that fixes the release script, and we now use Cocoapods 1.7.3


Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Internal/CoreRemoteMongoCollection.swift
```diff
@@ -398,14 +399,58 @@ public class CoreRemoteMongoCollection<T: Codable>: Closable {
                                                            withRequestTimeout: nil)
     }
 
+    /**
```
tkaye407: Totally fine with me, but why not make this all one function with optional default params

adamchel: I think this is the cleanest way to expose this given the current state of things. If we wanted it to be one function with optional params, it would be possible to pass in both a match filter and a list of IDs, which doesn't have a clear expected behavior. 

We could also pass in an enum of "watchType" where you have `.fullCollection`, `.matchFilter`, and `.idList`, but that would break the existing API for ID watch, and it could confuse people with the existing `fullDocument` vs. `compact` stream type in the id-based watch function.

Once we deprecate and remove compact watch and ID watch, we can clean this up and make it more Swifty. @jsflax feel free to weigh in as well

jsflax: Agreed on both sides. My only complaint would be that the position of the delegate arg switches, which is kind of just annoying.

jsflax: Keep in mind we can completely break out API once we move to Realm.

tkaye407: @adamchel makes sense to me, this is not a user-exposed function though? I have always kind of wondered why we tend to do this in the core modules

adamchel:
> This LGTM, but please test appending weak references to the streams array.

Discussed offline, we decided that the existing test `testStreamDoesCloseOnSwitch` in `CoreRemoteMongoCollectionUnitTests` satisfactorily covers this.

Core/Services/StitchCoreRemoteMongoDBService/Sources/StitchCoreRemoteMongoDBService/Internal/CoreRemoteMongoCollection.swift
adamchel: it's not user-exposed, but it maps pretty closely to what the end users see on the `RemoteMongoCollection` API. I didn't want to add extra complexity for an API we know we'll be changing in the future.

PR #197, opened 2019-11-23
STITCH-3632 add client sdk credentials for Apple sign in

jsflax:
- Add `AppleCredential` and `AppleAuthProvider` to `Auth` group
- Add apple as a `ProviderConfig` for Stitch Admin SDK

Note: Automated testing would be massive time sink (and likely incredibly brittle). I have tested this using Apple's example Apple Sign-in application. It works fine.


jsflax:
Yes, but we likely have to bring in the evergreen team to have the latest versions of Swift and Xcode.

PR #199, opened 2019-12-18
Swift SDK: Decode and expose `customData` on UserProfile

jsflax:


Core/StitchCoreSDK/Sources/StitchCoreSDK/Auth/Internal/CoreStitchUser.swift
```diff
@@ -46,6 +47,17 @@ public protocol CoreStitchUser {
...
     var lastAuthActivity: TimeInterval { get }
+
+    /**
+     You can store arbitrary data about your application users
+     in a MongoDB collection and configure Stitch to automatically
+     expose each user’s data in a field of their user object.
+     For example, you might store a user’s preferred language,
+     date of birth, or their local timezone.
+
+     If this value has not been configured, it will be empty.
```
mburdette: [nit] `if this functionality...` might be better than value?

Darwin/StitchCore/StitchCore/Core/Auth/Internal/StitchAuthImpl.swift
```diff
@@ -333,4 +333,10 @@ internal final class StitchAuthImpl: CoreStitchAuth<StitchUserImpl>, StitchAuth
             }
         }
     }
+
+    func refreshCustomData(_ completionHandler: @escaping (StitchResult<Void>) -> Void) {
```
mburdette: 👍 so this is public?

Darwin/StitchCore/StitchCore/Core/Auth/StitchUser.swift
```diff
@@ -62,6 +62,17 @@ public protocol StitchUser: CoreStitchUser {
...
     var lastAuthActivity: TimeInterval { get }
 
+    /**
+     You can store arbitrary data about your application users
+     in a MongoDB collection and configure Stitch to automatically
+     expose each user’s data in a field of their user object.
+     For example, you might store a user’s preferred language,
+     date of birth, or their local timezone.
+
+     If this value has not been configured, it will be empty.
```
mburdette: [nit] same here about `functionality` vs `value`

Darwin/StitchCore/StitchCoreTests/Core/StitchCustomUserDataIntTests.swift
```diff
@@ -0,0 +1,159 @@
...
+        XCTAssertEqual(user.customData, [:])
+
+        exp = expectation(description: "will refresh custom data")
+        client.auth.refreshCustomData({ _ in
+            exp.fulfill()
+        })
+        waitForExpectations(timeout: 10)
+
+        XCTAssertEqual(user.customData["favoriteColor"] as? String, "blue")
+
```
mburdette: [nit] since the `refreshCustomData` also refreshes an access token you might want to do a function call or something as this user to make sure the access token is updated

Darwin/StitchCore/StitchCore/Core/Auth/StitchUser.swift
jsflax: I actually copied this from the docs website, but I'll change it here.

Darwin/StitchCore/StitchCore/Core/Auth/Internal/StitchAuthImpl.swift
jsflax: Correct, this is exposed publicly through the protocol it's implementing.

